PR #7, opened 2017-11-07
Isssue #267 chore:Keycloak deployment and configurations

ahghatol:
…gin jar file


coolbung:
As discussed - remove the entire maven related code. If you run the build on the build slave, maven will already be installed.

rjshrjndrn:
closing the issue.

PR #31, opened 2018-02-05
Script to automate realm management user creation and initial realm import for new keycloak install

ahghatol:


ansible/roles/keycloak/templates/keycloak-bootstrap.conf.j2
```diff
@@ -0,0 +1,13 @@
+{
```
coolbung: Please put a comment for what each of these variables mean @ahghatol

ansible/roles/keycloak/files/python-keycloak-0.12.0/roles.json
```diff
@@ -0,0 +1,28 @@
+[
+    {
+        "id": "3ff462fc-b33c-431a-b54b-861c3298d910",
```
coolbung: is the id and containerId referenced anywhere else ?

ansible/roles/keycloak/defaults/main.yml
```diff
@@ -57,11 +57,11 @@ wildfly_standalone_config_path: /opt/keycloak/keycloak/standalone/configuration/
 keycloak_ext:  <extension module="org.keycloak.keycloak-server-subsystem"/>
 #keycloak_subsystem:
 #keycloak_cache_container:
-#keycloak_default_ds: 
+#keycloak_default_ds:
```
coolbung: You shouldn't be changing any code from the role files. These seem to be default variables that you can override when invoking the role.

PR #163, opened 2018-06-03
Issue #SB-3145 test: Added configuration file for testing

SMYALTAMASH:
Added Sunbird Post Installation Testing scripts


deploy/funcTest.sh
```diff
@@ -177,4 +177,5 @@ get_logs
 check_service_health
 
 #Zipping Logs
+sudo apt-get install zip -y
```
rjshrjndrn: Please use tar to compress, so that we don't have to install one more package.

PR #189, opened 2018-06-28
Issue #SB-4504 monitoring: Added Missing Variable and necessary Proxy…

SMYALTAMASH:
Added Missing Variable for monitoring to work and necessary Proxy configuration changes done.


ansible/roles/stack-monitor/defaults/main.yml
```diff
@@ -183,3 +183,5 @@ postgres_exporter_user: postgres_exporter
 
 root_group: root
 root_owner: root
+
+backup_storage_name: test
```
rjshrjndrn: Hey please give a name with some context, like sunbird-prometheus-backup or so

PR #245, opened 2018-09-20
Adding config file validation script, updating main install script to invoke validation, updated sanity script

msknext:
[SB-5443] - https://project-sunbird.atlassian.net/browse/SB-5443
Added a validation script to validate the config file.

[SB-7380] - https://project-sunbird.atlassian.net/browse/SB-7380
Removed tabs from config file as it is incorrect in yaml
Updated two variables in config file and removed spaces to maintain uniformity. This also doesn't give the correct result when we awk for the value.
Updated sanity script to correctly check the docker version


deploy/config
```diff
@@ -158,8 +158,8 @@ es_etc_node_datagroups['es']:
 es_api_host: "{{inventory_hostname}}"
 es_heap_size: 1500m
 es_etc_discovery_zen_minimum_master_nodes: 1
-es_snapshot_host: "{{ groups['es'][0] }}"
```
rjshrjndrn: @msknext why did you delete the spaces, as it's the best idiomatic ansible way to define variables.

deploy/validateConfig.sh
```diff
@@ -0,0 +1,172 @@
+#!/bin/bash
```
rjshrjndrn: can you have variables store all args, and then use the variables? It'll improve readability and maintainability

rjshrjndrn: please remove all `` and replace it with $()

msknext: @rjshrjndrn  Sorry I didn't get this point, you mean to create variables like implementation_name, database_host  etc and then get these values from the config file?
Yes ok `` will be replaced by $().

deploy/config
msknext: @rjshrjndrn  Yes I agree but two reasons:
1. To make it uniform with "{{inventory_hostname}}"
2. When doing an awk without -F ":", this will give the output as "{{.  As I saw in your original script you had used awk without -F ":", I used the same logic in my validate script to retrieve values into the array.
But it can be changed back . Please let me know.

deploy/validateConfig.sh
msknext: So what I have planned is to have all mandatory values in an associative array like below -
arr[env]]=$(awk '/env: /{ if ($2 !~ /#.*/) {print $2}}' config)
Then iterate over this array and based on key value invoke the validation function.
If I use just plain variables like below, -
env_value=$(awk '/env: /{ if ($2 !~ /#.*/) {print $2}}' config) - This would require me to have too many if else blocks to check each value and when trying the display the message, we have to manually echo the key each time like echo "Invalid env: value", echo "Invalid database_host: value".
An array would be easier. What do you think?

PR #280, opened 2019-01-11
updating cassandra default connection host

Kaali09:
By default, cassandra will reject requests from 127.0.0.1, but localhost will suffice.
Updating that


SMYALTAMASH:
Thanks for the contribution @Kaali09 this will solve restore issue

harshavardhanc:
You're awesome @Kaali09

PR #289, opened 2019-02-22
##fix added delay to fix the promethesu backup failure

anuthaharinimn:


ansible/roles/prometheus-backup/tasks/main.yml
```diff
@@ -1,5 +1,5 @@
 - name: ensure prometheus is stopped before backup
-  shell: "docker service scale monitor_prometheus=0"
+  shell: "docker service scale monitor_prometheus=0" && sleep 10
```
rjshrjndrn: I think it'll be proper to use ansible wait module. So that it'll be idempotent

PR #301, opened 2019-03-07
Release 1.15 sprint1: Rotate ssh keys for sunbird and diksha env

SMYALTAMASH:


ansible/roles/changekey/tasks/main.yml
```diff
@@ -0,0 +1,20 @@
...
+
+- name: Giving access to sudo without password
+  lineinfile:
+    path: /etc/sudoers
+    line: '{{bootstrap_user}} ALL=(ALL) NOPASSWD:ALL'
+    state: present
+
+- name: Remove existing authorized_keys from the "bootstrap_user" file and add a single key to its authorized_keys
+  become: yes
+  command: bash -lc "echo \"{{public_key}}\" > /home/{{bootstrap_user}}/.ssh/authorized_keys"
```
msknext: @SMYALTAMASH This will replace keys under bootstrap_user only. What about other user like ops? I would still be able to login as ops if I have the key since authorized_keys will be picked from /home/ops/.ssh?

PR #326, opened 2019-03-13
Issue #00 fix: adding new api whitelisting group

harshavardhanc:


rjshrjndrn:
use private varaibles in private group_vars.
You can use : 
```
- name: merging hash_a and hash_b into hash_c
  set_fact: hash_c="{{ hash_a|combine(hash_b) }}"
``` 
ref: https://stackoverflow.com/questions/35554415/in-ansible-how-to-combine-variables-from-separate-files-into-one-array

PR #331, opened 2019-03-14
added create user playbook and role

G33tha:


ansible/roles/add_user/tasks/main.yml
```diff
@@ -0,0 +1,15 @@
+- name: Add group
+  group: name="{{ ssh_user }}" state=present
+  become: yes
+
+- name:  Create user
+  become: yes
+  user: state=present name="{{ ssh_user }}" group="{{ ssh_user }}" groups=sudo shell=/bin/bash createhome=yes password="{{ password }}" update_password=always
```
msknext: @G33tha  This will add the user to sudo group. Is this intended? Do we want to add users to sudo group?

pipelines/ops/createuser/Jenkinsfile
```diff
@@ -0,0 +1,38 @@
...
+
+        ansiColor('xterm') {
+            stage('deploy'){
+                values = [:]
+                currentWs = sh(returnStdout: true, script: 'pwd').trim()
+                envDir = sh(returnStdout: true, script: "echo $JOB_NAME").split('/')[-3].trim()
+                module = sh(returnStdout: true, script: "echo $JOB_NAME").split('/')[-2].trim()
+                jobName = sh(returnStdout: true, script: "echo $JOB_NAME").split('/')[-1].trim()
+                ansiblePlaybook = "${currentWs}/ansible/createuser.yml"
+                ansibleExtraArgs = "--tags \"add-user\" --extra-vars \"remote=${params.instance_name} ssh_user=${params.username} password=${params.password} public_key=${params.public_key}\" --vault-password-file /var/lib/jenkins/secrets/vault-pass -v"
```
msknext: @G33tha Create a Jenkins drop down parameter with choices as add-user and delete-user. Take the value for tags as ${params.option}

ansible/roles/add_user/tasks/main.yml
G33tha: yes, we wanted add users to sudo group

PR #340, opened 2019-03-15
Issue #000 pr: ansible code for ml-workbench and jenkins pipeline files

SMYALTAMASH:


SMYALTAMASH:
We can have it then we should be using tags.

SMYALTAMASH:
@gandham-santhosh code changes are done accordingly to have a single ansible file for provision and deployment

PR #341, opened 2019-03-15
##fixing alert rules configuration for secor process monitoring

anuthaharinimn:
secor count for monitoring is defined as 8, but 9 secor process is running in secor vm, updated secor monitoring count variable to 9 to avoid false alerts. 

@rjshrjndrn Please review and merge.


rjshrjndrn:
@anuthaharinimn  please rebase the base to release-1.15

PR #358, opened 2019-03-20
Adding location delete api

manzarul:
Adding location delete api


SMYALTAMASH:
line number 3752 is having syntax issue it's not aligned with other api's @manzarul, please fix it
` - name: "getUserType"`

PR #396, opened 2019-04-08
Added the logstash role.

Kaali09:
This addition is for to provision the logstash
in cassandra nodes and read the audit log file 
and send to one of the kafka topic.


Kaali09:
sumbitting another pull request with some enhancement

PR #404, opened 2019-04-10
Added email and Slack alert for failed backup jobs. Fixed DD-110

santosh0705:
Added alerts for failing backup jobs.

### DIFF
--- a/pipelines/backup/cassandra-backup/Jenkinsfile
+++ b/pipelines/backup/cassandra-backup/Jenkinsfile
@@ -31,8 +31,20 @@ node() {
                 ansible_playbook_run(values)
             }
         }
+        currentBuild.result = 'SUCCESS'
     }
     catch (err) {
+        currentBuild.result = 'FAILURE'
+        slackSend (
+            channel: "${env.NOTIFY_SLACK_CHANNEL}",
+            color: 'danger',
+            message: "Build Failed - ${env.JOB_NAME} ${env.BUILD_NUMBER} (<${env.BUILD_URL}|Open>)"
+        )
         throw err
+    } finally {
+        step([$class: 'Mailer',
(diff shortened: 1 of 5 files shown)


pipelines/backup/cassandra-backup/Jenkinsfile
```diff
@@ -31,8 +31,20 @@ node() {
                 ansible_playbook_run(values)
             }
         }
+        currentBuild.result = 'SUCCESS'
     }
     catch (err) {
+        currentBuild.result = 'FAILURE'
+        slackSend (
+            channel: "${env.NOTIFY_SLACK_CHANNEL}",
```
msknext: @santosh0705 This would send slack notifications of all env's to same channel? Do you feel we need to have separate channels per env or all in once?
Slack alerts require a integration key. Do we have this already?
Also this is a global variable which isn't currently defined, please provide the value for this.

pipelines/backup/cassandra-backup/Jenkinsfile
```diff
@@ -31,8 +31,20 @@ node() {
...
+            color: 'danger',
+            message: "Build Failed - ${env.JOB_NAME} ${env.BUILD_NUMBER} (<${env.BUILD_URL}|Open>)"
+        )
         throw err
+    } finally {
+        step([$class: 'Mailer',
+            notifyEveryUnstableBuild: true,
+            recipients: "${env.NOTIFY_EMAIL_ADDRESS}",
```
msknext: Same - global variable which isn't currently defined, please provide the value for this. Again, this will send all env alerts.

santosh0705:
@msknext  This alert is for failed notification and only for interested audiences. Slack channel and email addresses are configurable either through global variable or job specific.

PR #406, opened 2019-04-10
Env variable changed for lms - SB:11172 (gzip impl)

iostream04:


SMYALTAMASH:
why is this variable used @iostream04 ?

PR #413, opened 2019-04-15
Issue #656: enabling CDN build

rjshrjndrn:
Building cdn assets while player deployment because of 
environment specific cdn urls.

### DIFF
--- a/pipelines/deploy/player/build-cdn.sh
+++ b/pipelines/deploy/player/build-cdn.sh
@@ -1,0 +1,21 @@
+#!/bin/bash
+# Build script
+# set -o errexit
+set -x
+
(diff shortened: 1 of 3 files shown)


pipelines/deploy/player/build-cdn.sh
```diff
@@ -0,0 +1,21 @@
...
+build_hash=$2
+artifact_version=$3
+npm install
+./node_modules/.bin/gulp download:editors
+cd client
+npm install
+npm run build-cdn -- --deployUrl $cdnUrl
+cd ..
+# Gzipping of assets
+./node_modules/.bin/gulp gzip:editors client:gzip
```
rjshrjndrn: @anoophm why do we need gzipping here ?

PR #495, opened 2019-05-27
Issue #000 player: cdn changed for player

SMYALTAMASH:


ansible/roles/stack-sunbird/tasks/service_stack.yml
```diff
@@ -2,14 +2,21 @@
 - name: Save stack file
   template: src="stack_{{service_name}}.yml" dest="/home/deployer/stack/stack_{{service_name}}.yml" mode=0644
 
+- name: Remove the existing config file
```
rjshrjndrn: you should have the file changed check. Because if we're doing simple deployment(without CDN for, say, variable changes) then we won't have index.js file, as we're cleaning the workspace. and we don't have to restart containers multiple times.

ansible/roles/stack-sunbird/tasks/service_stack.yml
```diff
@@ -2,14 +2,21 @@
...
+  shell: rm -rf /home/deployer/config/index_cdn.ejs
+  ignore_errors: yes
+
 - name: Save config file
   copy: src="{{inventory_dir}}/../../../sunbird-portal/src/app/dist/index_cdn.ejs" dest="/home/deployer/config/index_cdn.ejs" mode=0644
```
rjshrjndrn:
```yaml
copy:
  src: {{cdn_file_path}}
  dest: {{...
```
in Jenkins file
```sh
--extra-vars cdn_file_path=/home/dist/file.ejs
```

ansible/roles/stack-sunbird/tasks/service_stack.yml
rjshrjndrn: you can use something like this

ansible/roles/stack-sunbird/tasks/service_stack.yml
SMYALTAMASH: This is a good idea, I'll update this

ansible/roles/stack-sunbird/tasks/service_stack.yml
```diff
@@ -3,42 +3,43 @@
...
+
+- name: check if player is running
+  shell: docker service ps player_player
+  become: yes
+  register: playerRunning
   ignore_errors: yes
 
+- name: Remove the cdn file from  player if it's present
+  shell: docker service update --config-rm index_cdn.ejs player_player
+  when: copy_file.changed and playerRunning.changed and sunbird_portal_cdn_url is defined
```
rjshrjndrn: `when: copy_file.changed and sunbird_portal_cdn_url is defined and playerRunning.rc == 0`

PR #504, opened 2019-05-30
Feature #000 CDN: SUnbird Player CDN with Existing Config file logic

SMYALTAMASH:


SMYALTAMASH:
Tested with multiple scenarios and it's working as expected.
1> No CDN - Working
2> Redeploy No CDN - Working
3> Deploy with CDN - Working
4> Redeploy With CDN - Working
5> No CDN - Working

PR #532, opened 2019-06-10
SB-12982, SB-12983 User-org-service API onboarding

iostream04:
1. user  and org related API added for on boarding them on Dev server . 
2. URL's are as same as learner-service with minor changes, ex:  "v1.3" in place of "v1" .


iostream04:
@SMYALTAMASH  it is changed now

PR #592, opened 2019-07-03
ENC Service Deploynment ansible and JenkinsFile code

SMYALTAMASH:


ansible/roles/stack-sunbird/tasks/enc_service.yml
```diff
@@ -0,0 +1,9 @@
+---
+- name: Remove enc service
+  shell: "docker service rm enc-service"
+  ignore_errors: yes
+
+- name: Deploy user org service
+  shell: "docker service create --with-registry-auth --replicas {{ enc_replicas }} -p 9010:8013  --name enc-service --hostname enc-service --reserve-memory {{ enc_reservation_memory }} --limit-memory {{ enc_limit_memory }} --limit-cpu {{ enc_limit_cpu }} --reserve-cpu {{ enc_reservation_cpu }} --network application_default --env-file /home/deployer/env/sunbird_enc-service.env  {{hub_org}}/{{image_name}}:{{image_tag}}"
+  args:
+    chdir: /home/deployer/stack
```
keshavprasadms: @SMY-ALTAMASH why chdir when this is not going to use a stack file?

SMYALTAMASH: Chdir removed from the service file, it was added for testing purpose.
Application default network can be variablized but need to done for all the services and stacks in sunbird incluiding logging and monitoring. env file location will not change.

PR #609, opened 2019-07-11
Issue #000 fix: Security alert

rjshrjndrn:
Upgrading the package to use patched requests.

CVE-2018-18074
More information
moderate severity
Vulnerable versions: <= 2.19.1
Patched version: 2.20.0

The Requests package through 2.19.1 before 2018-09-14 for Python sends an HTTP Authorization header to an http URI upon receiving a same-hostname https-to-http redirect, which makes it easier for remote attackers to discover credentials by sniffing the network.


rjshrjndrn:
@ahghatol are we using this role. I think It's safe to delete the role, as we're now using subird-auth-deploy,

PR #628, opened 2019-07-15
Nginx optimizations

rjshrjndrn:
1. Upgrade nginx to 1.15.8
2. Support lua to expose metrics
3. Direct prometheus metrics scrapping
4. Limitting connections and increasing worker connections to mitigate slowloris

### DIFF
--- a/images/proxy/build.sh
+++ b/images/proxy/build.sh
@@ -1,5 +1,8 @@
 #!/bin/bash
 
+# Build script
+set -eo pipefail
+
 # Downloading deps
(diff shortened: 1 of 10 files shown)


images/proxy/build.sh
```diff
@@ -1,10 +1,21 @@
 #!/bin/bash
-# Build script
-set -eo pipefail
+
```
keshavprasadms: @rjshrjndrn Add this line `set -eo pipefail`

rjshrjndrn: :+1:

PR #649, opened 2019-07-22
added ansible role and jenkins file for prometheus federation

G33tha:


ansible/roles/prometheus-fed/defaults/main.yml
```diff
@@ -0,0 +1,22 @@
...
+enable_scraping_docker_metrics: false
+docker_metrics_port: "2377"
+monitor_stack_files_dest_dir: /opt/docker/stacks/prom_fed/stacks
+monitor_config_files_dest_dir: /opt/docker/stacks/prom_fed/config
+
+monitor_config_templates:
+  - prometheus.yml
+
+prometheus_route_prefix: prometheus
+prometheus_web_external_url: "{{proto}}://{{api__host}}:9090/{{ prometheus_route_prefix }}"
```
rjshrjndrn: please change the port

ansible/roles/prometheus-fed/templates/prometheus.yml
```diff
@@ -0,0 +1,45 @@
...
+  scrape_interval: 1m
+  scrape_timeout: 10s
+  evaluation_interval: 1m
+  # Clustername is defined for multi swarm cluster
+  # can name first cluster as DS1 and 2nd as DS2 ... DSn
+  # all metrics will have label as cluster: < value >
+  # This value won't be accessible from GUI but for external tools like
+  # federation or alertmanager
+  external_labels:
+    federation: fed1
```
rjshrjndrn: change the variable to cluster, so that we can have the same external label across all proms

ansible/roles/prometheus-fed/defaults/main.yml
```diff
@@ -0,0 +1,22 @@
+prometheus_reservation_memory: 1G
+prometheus_limit_memory: 1G
+prometheus_storage_retention_time: 90d
+
+# Defaults to 1KB. Good enough to catch errors mentioned in https://about.gitlab.com/2017/02/01/gitlab-dot-com-database-incident/
+expected_data_backup_size_in_bytes: 1024
```
rjshrjndrn: we can remove this, as we're not having any psql here

ansible/roles/prometheus-fed/templates/prometheus.yml
G33tha: as this is just doing a aggregation, we dont have to keep it as a cluster.

PR #740, opened 2019-08-16
Shared lib 2

Himanshu-optit:


vars/slack_notify.groovy
```diff
@@ -0,0 +1,34 @@
+def call() {
+       try {
+            ansiColor('xterm') {
+                String ANSI_GREEN = "\u001B[32m"
```
keshavprasadms: @Himanshu-optit These are not required
                String ANSI_GREEN = "\u001B[32m"
                String ANSI_NORMAL = "\u001B[0m"
                String ANSI_BOLD = "\u001B[1m"
                String ANSI_RED = "\u001B[31m"
                String ANSI_YELLOW = "\u001B[33m"

Remove them. They are already defined in Jenkins file

Himanshu-optit: Sure. I will make these changes

PR #919, opened 2019-11-19
Deploy: rename 'content-service' to 'knowledge-mw-service'

maheshkumargangula:


ansible/roles/stack-sunbird/tasks/knowledge-mw-service.yml
```diff
@@ -0,0 +1,9 @@
+---
+- name: Remove knowledge-mw service
+  shell: "docker service rm knowledge-mw-service"
+  ignore_errors: yes
+
```
SMYALTAMASH: We need to remove the old content-service which is running as well, for this change to take place

PR #929, opened 2019-11-21
Issue 0000 feat: Added desktop app blob url

anoophm:


SMYALTAMASH:
Good way of reusing existing variables @anoophm

PR #988, opened 2019-12-05
Kubernetes Rolling update for container deployments with zero downtime deployment of new services for all the core services.

SMYALTAMASH:


SMYALTAMASH:
Not required as of now

PR #1045, opened 2019-12-26
SC-1590

anmol2302:
**Ticket Ref:** SC-1590
1: Added /v1/user/exists API 
2: Added /v3/user/create

### DIFF
--- a/ansible/roles/kong-api/defaults/main.yml
+++ b/ansible/roles/kong-api/defaults/main.yml
@@ -5091,7 +5091,7 @@ kong_apis:
     - name: request-size-limiting
       config.allowed_payload_size: "{{ small_request_size_limit }}"
 
-  - name: userExistenceApi
+  - name: userExistenceApi_ACL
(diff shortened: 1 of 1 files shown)


ansible/roles/kong-api/defaults/main.yml
```diff
@@ -5090,3 +5090,22 @@ kong_apis:
...
+  - name: userExistenceApi
+    request_path: "{{ user_service_prefix  }}/v1/exists"
+    upstream_url: "{{ learning_service_url }}/v1/user/exists"
+    strip_request_path: true
+    plugins:
+    - name: jwt
+    - name: cors
+    - "{{ statsd_pulgin }}"
+    - name: acl
+      config.whitelist: "{{ getUserByKey_ACL | default(['appUpdate']) }}"
```
SMYALTAMASH: Give the same name as the API for the config.whitelist instead of getUserByKey_ACL give userExistenceApi_ACL

PR #1055, opened 2019-12-30
Kubernetes Offline installer changes to have proper redirection to download the lates artifact from  blob store

SMYALTAMASH:


SMYALTAMASH:
# duplicate

PR #1071, opened 2019-12-31
Readiness probe and liveness probe

G33tha:


kubernetes/helm_charts/core/adminutils/values.j2
```diff
@@ -22,6 +22,27 @@ strategy:
...
+adminutils:
+  readinessprobe:
+    enabled: true
+  livenessprobe:
+    enabled: true 
+  readinessProbe:
+    httpGet:
+      path: /health
+      port: 4000
+    initialDelaySeconds: 30
```
SMYALTAMASH: Variablize the timeout values with safe defaults for each service with a provision to override in private repo

rjshrjndrn: even for a single path overriding we've to override, entire dictionary. So I think variablizing only delaySeconds will add more variables.

kubernetes/helm_charts/core/content/values.j2
```diff
@@ -15,12 +15,33 @@ resources:
     cpu: {{content_cpu_limit|default('1')}}
     memory: {{content_mem_limit|default('1024Mi')}}
 network:
-  port: 9696
+  port: 9002
```
rjshrjndrn: check this port

PR #1086, opened 2020-01-03
issue #SC-1608 fix: changes for v3 user create api name to v1 signup

manzarul:
Changes for api name from user/v3/create to user/v1/signup


SMYALTAMASH:
PR is better now.

PR #1094, opened 2020-01-08
Helm charts for Assessment and Print service

rjshrjndrn:
private vs entry for print service
private helm nginx chart
fix: 
Automated env, conf configmap creation 
No need to have seperate tasks for env and .conf confimap creation

### DIFF
--- a/kubernetes/ansible/roles/sunbird-deploy/tasks/main.yml
+++ b/kubernetes/ansible/roles/sunbird-deploy/tasks/main.yml
@@ -4,33 +4,16 @@
     src: "{{ chart_path }}/values.j2"
     dest: "{{ chart_path }}/values.yaml"
 
-- name: copy env file from swarm role
+- name: copy and template env file from swarm role
   template:
     src: "{{service_env[release_name]}}"
-    dest: "{{role_path}}/templates/{{ release_name }}.env"
+    dest: "{{role_path}}/templates/"
 
-- debug: var=release_name
-  register: output
+- name: create env configmap
+  shell: find "{{role_path}}/templates/" -iname "*.env" -exec /bin/bash -c "ls {}; kubectl create configmap {{ release_name }}-config --from-env-file={} -n {{namespace}} --dry-run -o=yaml | kubectl apply -f - " \;
 
-- name: template vars
-  template:
-    src: "{{ release_name }}.env"
-    dest: "{{role_path}}/templates/{{ release_name }}.env"
-  when: output.release_name != "content"
-
-- name: template content service vars
-  template:
-    src: "{{ release_name }}.env"
-    dest: "{{role_path}}/templates/application.conf"
-  when: output.release_name == "content"
-
-- name: create configmap
-  shell: "kubectl create configmap {{ release_name }}-config --from-env-file={{role_path}}/templates/{{ release_name }}.env -n {{namespace}} --dry-run -o=yaml | kubectl apply -f -"
-  when: output.release_name != "content"
-
-- name: create content configmap
-  shell: "kubectl create configmap {{ release_name }}-config --from-file={{role_path}}/templates/application.conf -n {{namespace}} --dry-run -o=yaml | kubectl apply -f -"
-  when: output.release_name == "content"
+- name: create aplicaton configmap
+  shell: find "{{role_path}}/templates/" -iname "*.conf" -exec /bin/bash -c "ls {}; kubectl create configmap {{ release_name }}-config --from-file={} -n {{namespace}} --dry-run -o=yaml | kubectl apply -f - " \;
 
 - name: helm upgrade
   shell: helm upgrade --install --cleanup-on-fail {{ release_name }} {{ chart_path }} -n {{namespace}}
--- a/kubernetes/helm_charts/core/assessment/templates/configmap.yaml
+++ b/kubernetes/helm_charts/core/assessment/templates/configmap.yaml
@@ -1,0 +1,10 @@
+#apiVersion: v1
+#data:
+#  {{- range $key, $val := .Values.assessmentenv }}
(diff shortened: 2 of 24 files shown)


kubernetes/ansible/roles/sunbird-deploy/tasks/main.yml
```diff
@@ -4,33 +4,16 @@
...
-  when: output.release_name == "content"
-
-- name: create configmap
-  shell: "kubectl create configmap {{ release_name }}-config --from-env-file={{role_path}}/templates/{{ release_name }}.env -n {{namespace}} --dry-run -o=yaml | kubectl apply -f -"
-  when: output.release_name != "content"
-
-- name: create content configmap
-  shell: "kubectl create configmap {{ release_name }}-config --from-file={{role_path}}/templates/application.conf -n {{namespace}} --dry-run -o=yaml | kubectl apply -f -"
-  when: output.release_name == "content"
+- name: create aplicaton configmap
```
keshavprasadms: This will create conf file for other deployments also if no when condition is applied.

kubernetes/helm_charts/core/assessment/values.j2
```diff
@@ -0,0 +1,49 @@
...
+    cpu: {{assessment_cpu_req|default('50m')}}
+    memory: {{assessment_mem_req|default('50Mi')}}
+  limits:
+    cpu: {{assessment_cpu_limit|default('1')}}
+    memory: {{assessment_mem_limit|default('500Mi')}}
+network:
+  port: 9000
+strategy:
+  type: RollingUpdate
+  maxsurge: 1
```
keshavprasadms: Remove max.*. We are not using these in deployment manifests.

kubernetes/helm_charts/core/nginx-private-ingress/templates/deployment.yaml
```diff
@@ -0,0 +1,46 @@
...
+      app: private-ingress
+  template:
+    metadata:
+      labels:
+        app: private-ingress
+    spec:
+      containers:
+      - name: name
+        image: nginx
+        resources:
```
keshavprasadms: Add limits for resource

kubernetes/helm_charts/core/nginx-private-ingress/templates/deployment.yaml
```diff
@@ -0,0 +1,46 @@
...
+          requests:
+            cpu: 100m
```
keshavprasadms: Add readiness and liveliness probes - Check if nginx has a endpoint which gives 200

kubernetes/helm_charts/core/nginx-private-ingress/values.yaml
```diff
@@ -0,0 +1 @@
+namespace: dev
```
keshavprasadms: Delete this file

kubernetes/helm_charts/core/assessment/templates/configmap.yaml
```diff
@@ -0,0 +1,10 @@
+#apiVersion: v1
```
keshavprasadms: Changes to to same as print service configmap

kubernetes/helm_charts/core/print/values.j2
```diff
@@ -0,0 +1,49 @@
...
+replicaCount: {{print_replicacount|default(1)}}
+repository: {{print_repository|default('print-service')}}
+image_tag: {{ image_tag }}
+resources:
+  requests:
+    cpu: {{print_cpu_req|default('50m')}}
+    memory: {{print_mem_req|default('50Mi')}}
+  limits:
+    cpu: {{print_cpu_limit|default('1')}}
+    memory: {{print_mem_limit|default('1024Mi')}}
```
keshavprasadms: Set default to 500 Mi. Any reason to set min of 1Gb?

kubernetes/helm_charts/core/print/values.j2
```diff
@@ -0,0 +1,49 @@
...
+network:
+  port: 5000
+strategy:
+  type: RollingUpdate
+  maxsurge: 1
```
keshavprasadms: Remove max.*

kubernetes/pipelines/deploy_core/Jenkinsfile
```diff
@@ -2,8 +2,8 @@
 node() {
     try {
         stage('checkout public repo') {
-            cleanWs()
             checkout scm
+            sh'git clean -fxd'
```
keshavprasadms: Not exactly sure if we need to remove cleanws. This we will take it as an experiment and see how it goes.

kubernetes/ansible/roles/sunbird-deploy/tasks/main.yml
rjshrjndrn: nope. It'll create conf, only when there is a conf file, which we're copying for services from the swarm. More over we always copy one file.

kubernetes/helm_charts/core/assessment/values.j2
rjshrjndrn: But If we don't have this will kube wait for the container to be up, before killing the old one ?

kubernetes/helm_charts/core/print/values.j2
rjshrjndrn: it's running chrome inside. So it'll be good to have 1G limit

kubernetes/pipelines/deploy_core/Jenkinsfile
rjshrjndrn: cleanWs() is removed because always clone is taking more time. It makes more sense to pull the delta and do a git clean to remove previous templating or other changes.

PR #1095, opened 2020-01-08
Issue #00 fix: variablizing keycloak heap memory size

harshavardhanc:


ansible/roles/keycloak-deploy/templates/standalone.conf.j2
```diff
@@ -0,0 +1,73 @@
...
+# Uncomment the following line to prevent manipulation of JVM options
+# by shell scripts.
+#
+#PRESERVE_JAVA_OPTS=true
+
+#
+# Specify options to pass to the Java VM.
+#
+if [ "x$JAVA_OPTS" = "x" ]; then
+   JAVA_OPTS="-Xms64m -{{ keycloak_heap_size }} -XX:MetaspaceSize=96M -XX:MaxMetaspaceSize=256m -Djava.net.preferIPv4Stack=true"
```
keshavprasadms: MaxMetaspaceSize=256m 
Provide a variable for this - Set default value to as 256 only.

ansible/roles/keycloak-deploy/defaults/main.yml
```diff
@@ -77,6 +77,7 @@ keycloak_custom_ds: '
         </validation>
     </datasource>'
 
+keycloak_heap_size: Xmx1024m
```
keshavprasadms: Change default to 512m

harshavardhanc: Changes are updated

ansible/roles/keycloak-deploy/templates/standalone.conf.j2
harshavardhanc: Changes are updated

ansible/roles/keycloak-deploy/defaults/main.yml
```diff
@@ -77,6 +77,8 @@ keycloak_custom_ds: '
         </validation>
     </datasource>'
 
+keycloak_heap_size: Xmx512m
```
keshavprasadms: Since we don't want users making typo - Can we make this 512m and then add Xmx in the template?

harshavardhanc: Done

PR #1098, opened 2020-01-08
Issue #SC-1628 feat: Content-service logback configuration changes

PradyumnaNagendra:


ansible/roles/stack-sunbird/tasks/content-service.yml
```diff
@@ -16,6 +20,15 @@
   become: yes
   shell: "docker config create content-service.conf /home/deployer/config/content-service.conf"
 
+- name: Remove old logback.xml docker config
+  become: yes
+  shell: "docker rm /home/sunbird/content-service-1.0-SNAPSHOT/config/logback.xml"
```
keshavprasadms: @PradyumnaNagendra What are we doing here? docker rm is used to remove containers that are in stopped state and not config files.

ansible/roles/stack-sunbird/tasks/content-service.yml
```diff
@@ -16,6 +20,15 @@
...
+  ignore_errors: yes
+  
+- name: Copy new logback.xml
+  become: yes
+  shell: "docker cp /home/deployer/config/content-service_logback.xml /home/sunbird/content-service-1.0-SNAPSHOT/config/logback.xml"
```
keshavprasadms: This wont work - docker cp should be supplied with the conatiner name which is dynamic

PR #1104, opened 2020-01-08
Issue #00 fix: adding discovery.zen.minimum_master_nodes variable

harshavardhanc:


ansible/roles/es6/templates/elasticsearch.yml.j2
```diff
@@ -7,6 +7,12 @@
 cluster.name: elasticsearch
 {% endif %}
 
+{% if (groups['es']|length) == 1 %}
```
keshavprasadms: Update to length <= 2

PR #1215, opened 2020-02-10
Release 2.6.0

keshavprasadms:


kubernetes/ansible/roles/sunbird-monitoring/templates/prometheus-operator.yaml
```diff
@@ -92,7 +92,7 @@ routes:
             to: '{{ filter.alerts_mailing_list }}'
             html: '{% raw %}{{ template "email.default.html" . }}{% endraw %}'
             headers:
-              subject: '[{{ env }}-{{cluster_name}}] {% raw %}{{ .GroupLabels.alertname }}{% endraw %}'
+              subject: '[{{ env }}-KUBERNETES] {% raw %}{{ .GroupLabels.alertname }}{% endraw %}'
```
rjshrjndrn: we should use the cluster_name vairable and then default value kubernetes.
like this. https://github.com/rjshrjndrn/sunbird-devops/blob/release-2.8.0/kubernetes/ansible/roles/sunbird-monitoring/templates/prometheus-operator.yaml
Here the variable is kubernetes_cluster_name, because cluster_name variable is already getting overriden in hosts, and as of now, hosts is a softlink

PR #1267, opened 2020-02-27
SB-18019 added cert search and read API

anmol2302:
**Ticket Ref:** [SB-18019](https://project-sunbird.atlassian.net/browse/SB-18019?filter=10743)
1: Added `certreg/v1/certs/read`
2: Added `certreg/v1/certs/search`


SMYALTAMASH:
Do We really need these API's, can't the portal call them internally?

anmol2302:
> Do We really need these API's, can't the portal call them internally?

@SMY-ALTAMASH yes

PR #1268, opened 2020-02-27
fix: keycloak templating and Jenkinsfile fix

keshavprasadms:


ansible/roles/keycloak-deploy/templates/standalone-ha.xml
```diff
@@ -228,20 +239,20 @@
...
                 <local-cache name="users">
                     <object-memory size="10000"/>
                 </local-cache>
-                <distributed-cache name="sessions" owners="1"/>
-                <distributed-cache name="authenticationSessions" owners="1"/>
-                <distributed-cache name="offlineSessions" owners="1"/>
-                <distributed-cache name="clientSessions" owners="1"/>
-                <distributed-cache name="offlineClientSessions" owners="1"/>
-                <distributed-cache name="loginFailures" owners="1"/>
+                <distributed-cache name="sessions" owners="2"/>
```
SMYALTAMASH: @amit-tarento should this value be always 2 irrespective of the keycloak servers or it is based on the number of servers making up a cluster?

amit-tarento: Change the owner value to 1 as per existing config from stage keycloak configuration, As we made this changes for session testing on dev env.

SMYALTAMASH: @keshavprasadms please update owner value as 1

PR #1299, opened 2020-03-12
SB-14588 - Content Upload API refactoring - adding config

amitpriyadarshi:


amitpriyadarshi:
These properties are added with respect to storage-service. These config properties are not new but migrated from learning-service as part of upload api.

PR #1311, opened 2020-03-19
updated readiness and liveness timout seconds for content service

G33tha:


ansible/roles/stack-sunbird/tasks/content-service.yml
```diff
@@ -30,6 +30,6 @@
...
+  shell: "docker service create --with-registry-auth --replicas {{ content_service_replicas }} -p 9002:9000  --name content-service --hostname content-service --reserve-memory {{ content_service_reservation_memory }} --limit-memory {{ content_service_limit_memory }} --limit-cpu {{ content_service_limit_cpu }} --reserve-cpu {{ content_service_reservation_cpu }} --health-cmd 'wget -qO- content-service:9000/health || exit 1' --health-timeout 10s --health-retries 3  --network application_default --config source=content-service.conf,target=/home/sunbird/content-service-1.0-SNAPSHOT/config/application.conf,mode=0644 --config source=content-service_logback.xml,target=/home/sunbird/content-service-1.0-SNAPSHOT/config/logback.xml,mode=0644  {{hub_org}}/{{image_name}}:{{image_tag}}"
```
keshavprasadms: Change retry to 5 as we are trying 5 times in Kube.

ansible/roles/stack-sunbird/defaults/main.yml
```diff
@@ -358,6 +358,9 @@ content_liveness_readiness:
       port: 9000
     initialDelaySeconds: 60
     periodSeconds: 10
+    timeoutSeconds: 5
+    failureThreshold: 5
+    successThreshold: 1
```
keshavprasadms: This is the default value, so no need to define successThreshold

ansible/roles/stack-sunbird/defaults/main.yml
```diff
@@ -349,7 +349,7 @@ content_liveness_readiness:
       port: 9000
     initialDelaySeconds: 30
     periodSeconds: 10
-    timeoutSeconds: 5
+    timeoutSeconds: 10
     failureThreshold: 5
     successThreshold: 1
```
keshavprasadms: This is the default value, so no need to define successThreshold

ansible/roles/stack-sunbird/tasks/content-service.yml
G33tha: updated retry to 5 times

PR #1330, opened 2020-03-29
New Standards For Sunbird API STACK With Revised APIS and ACLS.

SMYALTAMASH:


rjshrjndrn:
@SMY-ALTAMASH,
 please squash all duplicate comments

ansible/roles/kong-api/defaults/main.yml
```diff
@@ -126,7 +126,7 @@ kong_apis:
         - 'appAccess'
     - name: rate-limiting
       config.policy: local
-      config.hour: "{{ medium_rate_limit_per_hour }}"
+      config.hour: "{{ acceptTermsAndCondition_limit_per_hour if acceptTermsAndCondition_limit_per_hour else medium_rate_limit_per_hour }}"
```
rjshrjndrn: why can't we just have a `{{ var | d() }}` rather than the if else. So that it'll be idiomatic and easy to understand

SMYALTAMASH: i think you saw an older commit @rjshrjndrn

SMYALTAMASH: It is no more valid now

kubernetes/ansible/roles/kong-api/tasks/main.yml
```diff
@@ -12,9 +12,5 @@
 - name: Copy kong api scripts
   copy: src=static-files/kong-api-scripts dest=/tmp mode=0755
```
rjshrjndrn: We've to have a check whether that port is accessible or not.

SMYALTAMASH: this task file is no more used for kubernetes in release-2.10.0 @rjshrjndrn , it is a different role altogether kubernetes/ansible/roles/kong-api/

rjshrjndrn: then which role is used to onboard APIs ? ansible/role ?. I think it might be good to add wait_for to check the port in the file, because most probably, we're going to use this file in future.

SMYALTAMASH: kubernetes/ansible/roles/kong-api/tasks/main.yml is the file which is used and this defaults/main.yml is being symblinked

SMYALTAMASH: kubernetes/ansible/roles/kong-api/tasks/main.yml is the role yml which will be used and kubernetes/ansible/api-manager.yml is the yml being used and ansible/roles/kong-api/defaults/main.yml is symblinked.

PR #1353, opened 2020-04-06
Release 2.9.0 to 2.10.0

keshavprasadms:


ansible/roles/kong-api/defaults/main.yml
```diff
@@ -157,6 +157,108 @@ kong_apis:
     - name: request-size-limiting
       config.allowed_payload_size: "{{ small_request_size_limit }}"
 
+  - name: PrivateContentRetireAPI
```
SMYALTAMASH, line 160: We don't need these api's anymore @keshavprasadms

PR #1355, opened 2020-04-06
fix: process exporter updates and fixes

keshavprasadms:
- Labelling the node to run process exporter prom
- removed gather_facts
- removed handlers
- removed pid for processes
- removed wal compression for prometheus


rjshrjndrn:
@keshavprasadms this seperate prometheus is supposed to be a temporary thing, right? till we finish our testng.

keshavprasadms:
> @keshavprasadms this seperate prometheus is supposed to be a temporary thing, right? till we finish our testng.

This will live tillfed / stateful lives I guess.

PR #1361, opened 2020-04-09
fix: updated dashboards, kibana chart, nginx ingress, helm deploy

keshavprasadms:
- feat to run kibana in kubernetes as a service witthout swarm
- kibana helm chart fixes and removal of unnessary files
- helm deploy ansible updates to check deployments, daemonsets and statefulsets
- nginx private ingress standardization as per other charts
- added nginx and apimanger dashboard for kubernetes and updated process exporter dashboard
- updated helm chart of apimanger for dns caching
- updated kibana, fluent-bit and oauth2-proxy helm charts to override the public charts


kubernetes/helm_charts/logging/kibana/templates/deployment.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: apps/v1
 kind: Deployment
 metadata:
-  name: {{ template "fullname" . }}
+  name: kibana
```
rjshrjndrn, line 4: why are we changing public helm charts? it'll be difficult to upgrade, right. can't we override this using `-f values.yaml`

rjshrjndrn: you can have a look in sunbird-monitoring role in k8s

keshavprasadms: We will not use the public helm chart for this. We will keep this internal only as these changes are very rare. When we do, we have a bigger worry to do ES first than Kibana,

PR #1406, opened 2020-04-20
added acl for report service

G33tha:


ansible/roles/kong-consumer/defaults/main.yml
```diff
@@ -100,6 +100,9 @@ kong_all_consumer_groups:
   - domainUpdate
   - domainUser
   - privateContentMgmt
+  - reportsAccess
+  - reportsUpdate
+  - reportsCreate
```
SMYALTAMASH, line 105: Reports admin is also required

G33tha: report admin is also added

PR #1421, opened 2020-04-23
Issue #001 fix: updating host group name

raghupathiguduri:
Updating host group name as  we have group with same name.


rjshrjndrn:
Not needed

PR #1469, opened 2020-05-13
Issue #00 fix: log-es clustering and helm chart for fluentd

harshavardhanc:
Ansible changes for log-es clustering - updated host name to point to log-es group and the cluster name changes to make the cluster name same across all the nodes.
Helm chart for fluentd, fluentd will point to all the es nodes in the cluster.
Kibana changes to point to all the es nodes in the cluster.


kubernetes/helm_charts/logging/kibana/templates/deployment.yaml
```diff
@@ -79,15 +79,6 @@ spec:
...
         imagePullPolicy: "{{ .Values.imagePullPolicy }}"
         env:
-          {{- if .Values.elasticsearchURL }}
-          - name: ELASTICSEARCH_URL
-            value: "{{ .Values.elasticsearchURL }}"
-          {{- else if .Values.elasticsearchHosts }}
-          - name: ELASTICSEARCH_HOSTS
-            value: "{{ .Values.elasticsearchHosts }}"
-          {{- end }}
-          - name: SERVER_HOST
```
rjshrjndrn, line 89: add a if-else block and raise a PR. Put the pr here as comment. In future when we upgrade the helm chart, we should be overwrite it with upstream without any manual change

harshavardhanc: @rjshrjndrn This changes are specifically for kibana to point to multiple es nodes. If it has hosts in env it will read it here instead of that we are passing it in kibana config file.

rjshrjndrn: @harshavardhanc for that we can keep the hosts as empty. and in values.yaml we can override SERVER_HOST as 0.0.0.0

harshavardhanc: @rjshrjndrn this is done

PR #1474, opened 2020-05-14
Issue #SC-0000 feat: Added logback and relaxed kafka appender

krgauraw:


ansible/roles/stack-sunbird/tasks/assessment-service.yml
```diff
@@ -7,6 +7,10 @@
   become: yes
   template: src=assessment-service_application.conf dest=/home/deployer/config/assessment-service.conf mode=0644
 
+- name: Save logback configuration into a env file
```
SMYALTAMASH, line 10: These are for swarm and it will not be required anymore @krgauraw

krgauraw: As discussed, this will not be added for any other service.

PR #1509, opened 2020-05-23
##fix: issue - adding private-Api in all group

raghupathiguduri:


ansible/roles/kong-consumer/defaults/main.yml
```diff
@@ -43,6 +43,7 @@ kong_all_consumer_groups:
   - contentAdmin
   - contentCreate
   - contentUpdate
+  - contentSuperAdmin
```
SMYALTAMASH, line 46: we should never give access to private apis to the appliaction, it is against the standards

PR #1518, opened 2020-05-27
Issue #001: kafka and secor var update

GangaramD:
Kafka lag varible addition to include the varibles which will override the default value from the private repo and secor job count update to 15.


kubernetes/helm_charts/monitoring/alertrules/templates/promrulesProcess.yml
```diff
@@ -53,7 +53,7 @@ spec:
         message: {{`'Number of running processes are: {{$value}}'`}}
         summary: More than one process running
     - alert: secor_process_not_running_critical
-      expr: namedprocess_namegroup_states{groupname="secor",state="Sleeping"} != 13
+      expr: namedprocess_namegroup_states{groupname="secor",state="Sleeping"} != 15
```
rjshrjndrn, line 56: Make the count as a variable and update that in private repo

GangaramD: updated the changes.

PR #1556, opened 2020-06-05
Issue #SB TG-60 fix: offline keycloak token validation

anoophm:


ansible/roles/stack-sunbird/templates/sunbird_knowledge-mw-service.env
```diff
@@ -56,3 +56,5 @@ sunbird_cassandra_consistency_level={{sunbird_cassandra_consistency_level}}
 {% endif %}
 # Release-1.15
 sunbird_health_check_enable={{sunbird_health_check_enable}}
+# Release-3.0.0
+sunbird_keycloak_public_key={{sunbird_keycloak_public_key}}
```
rjshrjndrn, line 60: can this variable be empty ? in that case we've to have `|d('')`
else you'll have to add this variable in `https://github.com/project-sunbird/sunbird-devops/blob/master/private_repo/ansible/inventory/dev/Core/secrets.yml`

PR #1557, opened 2020-06-05
Issue #DP-427 fix: Reset password and redirect to login page from back to application

JagadishPujari:
[JIRA LINK](https://project-sunbird.atlassian.net/browse/DP-427)


ansible/artifacts/sunbird/login/resources/js/login.js
```diff
@@ -395,6 +398,16 @@ var forgetPassword = (redirectUrlPath) => {
 	}
 }
 
+var backToApplication = () => {
+	var redirect_uri = getValueFromSession('redirect_uri');
+	if (redirect_uri) {
+		var updatedQuery = sessionStorage.redirect_uri.split('?')[0];
+		window.location.href = updatedQuery;
+	}else{
```
vinukumar-vs, line 406: redirect_uri will be always present. It will never go to else block.

ansible/artifacts/sunbird/login/info.ftl
```diff
@@ -12,9 +12,9 @@
                     <#if skipLink??>
                     <#else>
                         <#if pageRedirectUri??>
-                            <p><a href="${pageRedirectUri}">${msg("backToApplication")}</a></p>
+                            <p><a onclick="javascript:backToApplication()">${msg("backToApplication")}</a></p>
```
vinukumar-vs, line 15: We should not put any checks here. What ever the base or redirect uri is present. It should go there. Because this is generic page for any error accored by keycloak.

PR #1559, opened 2020-06-05
Chatapi config

G33tha:


kubernetes/helm_charts/core/nginx-public-ingress/values.j2
```diff
@@ -622,6 +622,20 @@ proxyconfig: |
       proxy_http_version 1.1;
       proxy_pass http://kong;
     }
+    
+    location ~ /chatapi/bot {
+      rewrite ^/chatapi/bot(.*) /bot$1 break;
```
rjshrjndrn, line 627: I think it'll be better to have
```
rewrite ^/chatapi/(.*) /$1 break;
```

G33tha: @rjshrjndrn  rewrite url updated

PR #1567, opened 2020-06-09
feat: adding code post loadtest fixes

keshavprasadms:
- Kong JWT Plugin change
- Kong 0.12.3 build and deploy scripts
- Adminutils mount of rsa keys
- Kong JWT onboarding for v2 endopint in kong
- Generic role to mount keys to any service
- Kong API syntax changes for 0.12.3 version compatibilty
- New v2 consumer for mobile app


kubernetes/ansible/roles/helm-deploy/tasks/main.yml
```diff
@@ -14,9 +14,24 @@
...
+  vars:
+    private_key_path: "{{ kong_private_key_path | default ('keys') }}"
+    private_key_sign_start: "{{ kong_private_key_sign_start | default(0) }}"
+    private_key_sign_end: "{{ kong_private_key_sign_end | default(1) }}"
+  when: release_name == "adminutils" and adminutil__jwt_keycount is defined and adminutil__jwt_keycount > 0
+
 - name: helm install
   shell: helm upgrade --install --atomic --timeout 10m {{ release_name }} {{ chart_path }} -n {{namespace}}
 
+- name: Remove the keys directory post helm install
```
rjshrjndrn, line 29: this task will never get executed if the above installation fails for any reason. This will be good for debugging the secrets. But I don't think helm deployment will fail because of secrets. Do we need to add `final` block ?

keshavprasadms: @rjshrjndrn Deployment will fail if keys are incorrect or in different format. Adminutil will crash loop.

rjshrjndrn: oh.. Okay. Then make sense to keep the decrypted files, as it'll help to debug

PR #1576, opened 2020-06-12
feat: Code base update for groot

keshavprasadms:
- Generic role update
- Sunbird And Helm deploy role updates for key mounts to adminutil, lms and learner
- General fixes


kubernetes/ansible/roles/mount-keys/tasks/encrypt.yml
```diff
@@ -0,0 +1,6 @@
+---
```
rjshrjndrn, line 1: why do we need  this ?

keshavprasadms: Its yaml identifier right.

rjshrjndrn: Not that... The play book ... Encrypt again?

keshavprasadms: Its used to encrypt the private keys after generating public keys from the decrypted private keys.

rjshrjndrn: so we don't need this anymore right

keshavprasadms: Another option would be to just copy the data from private back to inventory directory which would be encrypted. Or just remove the keys directory post install from inventory directory that way we can save time not to encrypt again. In case of debugging, we would need to copy the keys directory manually again and decrypt if we remove the directory.

keshavprasadms: Please let me know what I should change to. Do we copy keys directory again just for the sake of original inventory copied structure or remove the directory

keshavprasadms: Changed to delete model

PR #1577, opened 2020-06-12
Issue #SH-397 adding access-token properties to learner-service env

Hari-stackroute:
This pr is regarding adding access token which helps in accessing adminutil api and decrypting the public tokens.
We need to add accesstoken.publickey.basepath value as path: /keys  inside the container.


SMYALTAMASH:
@Hari-stackroute @keshavprasadms , can we close this PR and raise a new one to latest release branch?

PR #1584, opened 2020-06-15
ISSUE#SB-19521: added new env vars (private container) for print service

aishwa8141:
print-service uses both public and private container
the private container is used for certificates


SMYALTAMASH:
Looks Good to me

PR #1585, opened 2020-06-15
Added configuration value fo qrcode courses limit

rheafernandes:


ansible/roles/stack-sunbird/templates/sunbird_lms-service.env
```diff
@@ -117,3 +117,4 @@ managed_user_limit={{managed_user_limit | default(30)}}
 
 adminutil_base_url = {{adminutil_base_url}}
 adminutil_sign_endpoint = {{adminutil_sign_endpoint | default('v1/sign/payload')}}
+sunbird_user_qrcode_courses_limit=5000
```
rjshrjndrn, line 120: we should have a variable to override this and add 5000 as default @maheshkumargangula

rheafernandes: @rjshrjndrn  resolved

PR #1599, opened 2020-06-19
program/v1/tenat/list API changes added

twsvaishali:


ansible/roles/kong-api/defaults/main.yml
```diff
@@ -4429,3 +4429,20 @@ kong_apis:
...
+  - name: programTenantList
+    request_path: "{{ program_service_prefix }}/v1/tenant/list"
+    upstream_url: "{{ program_service_url }}/program/v1/tenant/list"
+    strip_request_path: true
+    plugins:
+    - name: jwt
+    - name: cors
+    - "{{ statsd_pulgin }}"
+    - name: acl
+      config.whitelist: 'userUpdate'
```
SMYALTAMASH, line 4442: it is having a wrong ACL @harshavardhanc @twsvaishali what is the intent of this API?

PR #1614, opened 2020-06-24
Updated configuration for disabling backward compatibility for content tagging properties

rheafernandes:
…t tagging properties

### DIFF
--- a/ansible/roles/stack-sunbird/templates/search-service_application.conf
+++ b/ansible/roles/stack-sunbird/templates/search-service_application.conf
@@ -346,4 +346,4 @@ compositesearch.index.name="{{ compositesearch_index_name }}"
 
 content.tagging.backward_enable=true
 content.tagging.property=["subject","medium"]
-search.payload.log_enable=true
+search.payload.log_enable=false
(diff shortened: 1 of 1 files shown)


ansible/roles/stack-sunbird/templates/search-service_application.conf
```diff
@@ -346,4 +346,4 @@ compositesearch.index.name="{{ compositesearch_index_name }}"
 
 content.tagging.backward_enable=true
 content.tagging.property=["subject","medium"]
-search.payload.log_enable=true
+search.payload.log_enable=false
```
rjshrjndrn, line 349: I think this should be variable. 
```
 search.payload.log_enable={{search_payload_log_enabled|d(false)}}
```

amitpriyadarshi: Hi Rajesh, it is wrong property. THis property was not expected to change. It was just confusion.

rheafernandes: It was wrong property, sorry.

PR #1639, opened 2020-07-03
groups service creationa and automation

SMYALTAMASH:


SMYALTAMASH:
Groups Microservice Creation and deployment scripts with ansible automation, Kubernetes manifests and helms charts creation with health liveliness and readiness probe.

PR #1641, opened 2020-07-03
fix: Don't send Bad Token for null / empty Bearer tokens

keshavprasadms:


images/kong/plugins/0.12.3/jwt/handler.lua
```diff
@@ -114,7 +114,9 @@ local function do_authentication(conf)
   -- Decode token to find out who the consumer is
   local jwt, err = jwt_decoder:new(token)
   if err then
-    return false, {status = 401, message = "Bad token; " .. tostring(err)}
+    -- Don't sent Bad Token for null / empty Bearer tokens
+    -- return false, {status = 401, message = "Bad token; " .. tostring(err)}
```
rjshrjndrn, line 118: remove this commented line @keshavprasadms

keshavprasadms: @rjshrjndrn Removed that. Also changed kong tag.

PR #1671, opened 2020-07-10
Issue #SB-15992 fix: Added env variable for phrase app token

Nik720:


ansible/roles/stack-sunbird/templates/sunbird_player.env
```diff
@@ -102,3 +102,4 @@ sunbird_bot_service_URL=/chatapi/bot
 sunbird_p1_reCaptcha_enabled={{sunbird_p1_reCaptcha_enabled|default('true')}}
 sunbird_p2_reCaptcha_enabled={{sunbird_p2_reCaptcha_enabled|default('true')}}
 sunbird_p3_reCaptcha_enabled={{sunbird_p3_reCaptcha_enabled|default('true')}}
+sunbird_phraseApp_token={{sunbird_phraseApp_token|''}}
```
keshavprasadms, line 105: This is incorrect. Please update with default keywork

Nik720: Fixed

rjshrjndrn: what will happen if the variable is empty ? @Nik720

Nik720: If this token is empty then it will call the old method.

keshavprasadms: Resolved

PR #1673, opened 2020-07-11
Issue #SH-587 added user-existence api v2

Hari-stackroute:


ansible/roles/kong-api/defaults/main.yml
```diff
@@ -1037,6 +1037,24 @@ kong_apis:
     - name: request-size-limiting
       config.allowed_payload_size: "{{ small_request_size_limit }}"
 
+  - name: userExistenceApiV2
```
keshavprasadms, line 1040: Please add it as per alphabet order and not random

Hari-stackroute: @keshavprasadms : added according to alphabet-order

PR #1701, opened 2020-07-14
added new route for mvc content create api

tanmayranjan:


ansible/roles/kong-api/defaults/main.yml
```diff
@@ -5104,6 +5104,24 @@ kong_apis:
...
+    request_path: "{{ mvc_service_prefix }}/v3/create"
+    upstream_url: "{{ mvc_service_url }}/v3/mvccontent/create"
+    strip_request_path: true
+    plugins:
+    - name: jwt
+    - name: cors
+    - "{{ statsd_pulgin }}"
+    - name: acl
+      config.whitelist: 
+        - 'contentAdmin'
```
harshavardhanc, line 5117: change the whitelist name to contentCreate instead of contentAdmin

PR #1749, opened 2020-07-22
fix: merge 3.0.1 to 3.1.0

keshavprasadms:


kubernetes/ansible/roles/sunbird-monitoring/defaults/main.yml
```diff
@@ -13,6 +13,7 @@ monitoring_stack:
   - kafka-topic-exporter
   - prometheus-redis-exporter
   - kafka-exporter
```
rjshrjndrn, line 15: Please remove this as this helm chart is not present

PR #1781, opened 2020-08-06
Issue #HE-8 feat: env var for deactivate job api host in report service.

yravinderkumar33:


ansible/roles/stack-sunbird/templates/sunbird_report-service.env
```diff
@@ -8,3 +8,4 @@ SUNBIRD_BASE_REPORT_URL=report
 SUNBIRD_REPORTS_TABLE_NAME=report
 SUNBIRD_REPORT_SUMMARY_TABLE_NAME={{sunbird_report_summary_table_name}}
 SUNBIRD_ENV="https://{{domain_name}}"
+DEACTIVATE_JOB_API_HOST=http://{{analyticsapi_ip}}:8000/data/v1/report/jobs/deactivate
```
SMYALTAMASH, line 11: analytics API runs in port 9000 or 8000 @yravinderkumar33 ?

SMYALTAMASH: sunbird_analytics_api_base_url use this variables instead.

SMYALTAMASH: check with internal team as to what should be the correct endpoint for this, i believe the value is wrong. If you're still pointing to analytics API then use the sunbird_analytics_api_base_url variable.

PR #1794, opened 2020-08-10
Merge Release 3.0.0 to 3.0.1

keshavprasadms:


kubernetes/helm_charts/monitoring/ingestion-kafka-exporter/templates/servicemonitor.yaml
```diff
@@ -0,0 +1,33 @@
+{{- if .Values.prometheus.serviceMonitor.enabled }}
```
rjshrjndrn, line 1: @Kaali09 why are we having a separate chart? we were supposed to have a soft link right ?

Kaali09: @rjshrjndrn we are using .Chart.Name variable in deployment manifest . chart.yaml different for both .

rjshrjndrn: please override it in values as `nameOverride` can override the value.

PR #1811, opened 2020-08-12
Issue #TG-465 Nginx changes

rjshrjndrn:
1. trace id propagation
2. Automatic config reload
3. channel, device, app Id logging
4. Content caching

### DIFF
--- a/kubernetes/helm_charts/core/nginx-public-ingress/templates/daemonset.yaml
+++ b/kubernetes/helm_charts/core/nginx-public-ingress/templates/daemonset.yaml
@@ -4,6 +4,8 @@ kind: DaemonSet
 metadata:
   name: nginx-public-ingress
   namespace: {{ .Values.namespace }}
+  annotations:
+    nginxRolloutID: {{ randAlphaNum 5 | quote }} # Restart nginx after every deployment
 spec:
   selector:
     matchLabels:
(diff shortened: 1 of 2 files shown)


kubernetes/helm_charts/core/nginx-public-ingress/templates/daemonset.yaml
```diff
@@ -4,8 +4,6 @@ kind: DaemonSet
 metadata:
   name: nginx-public-ingress
   namespace: {{ .Values.namespace }}
-  annotations:
```
keshavprasadms, line 7: Why are we removing skater annotation?

kubernetes/helm_charts/core/nginx-public-ingress/templates/daemonset.yaml
```diff
@@ -5,7 +5,8 @@ metadata:
   name: nginx-public-ingress
   namespace: {{ .Values.namespace }}
   annotations:
-    reloader.stakater.com/auto: "true"
+    # Restarting nginx if ssl cert is changed
+    secret.reloader.stakater.com/reload: "ingress-cert"
```
keshavprasadms, line 9: Please add merge domain secrets also

keshavprasadms (changes requested):
Please add merge domain secrets also

kubernetes/helm_charts/core/nginx-public-ingress/templates/daemonset.yaml
rjshrjndrn: Removing the entire block. Nginx reload will take care of the ssl refresh also

keshavprasadms (approved):

PR #1819, opened 2020-08-17
Issue #0000 feat: TG348 added the fluentd exlcuded conf

Kaali09:


kubernetes/ansible/roles/logging/templates/fluentd-elasticsearch.yaml
```diff
@@ -48,3 +48,82 @@ extraConfigMaps:
       </buffer>
     </match>
     </label>
+
+  containers.input.conf: |-
```
keshavprasadms, line 52: You have to disable the default configmap in the dictionary  before this gets applied. Check this line

PR #1823, opened 2020-08-19
SB-19349: point downloadCertificate api to cert-registry

aishwa8141:
For download cert, the learner-service holds the endpoint for older mobile clients.
so changing the kong configuration to point to cert-registry rather than learner-service


indrajra (approved):
{{ user_service_prefix }}/v1/certs/download is still retained, whereas only the upstream is changed here. Thanks!

keshavprasadms (approved):

PR #1827, opened 2020-08-25
variablizing nginx worker process, keep default to 2

rjshrjndrn:


kubernetes/helm_charts/core/nginx-public-ingress/values.j2
```diff
@@ -678,7 +678,7 @@ proxyconfig: |
   }
 nginxconfig: |
   user  nginx;
-  worker_processes  1;
+  worker_processes  {{nginx_worker_processes | d(2)}};
```
keshavprasadms, line 681: @rjshrjndrn Use `auto` instead of defaulting to 2

keshavprasadms (changes requested):
Please see comments

keshavprasadms (approved):

PR #1893, opened 2020-09-16
Issue #SB-20637 feat:fixed controller logging

amit-tarento:
Fixed controller logging

### DIFF
--- a/ansible/roles/stack-sunbird/templates/learner-service_logback.xml
+++ b/ansible/roles/stack-sunbird/templates/learner-service_logback.xml
@@ -6,8 +6,25 @@
   <timestamp key="timestamp" datePattern="yyyy-MM-dd"/>
   <!-- common transactions logs -->
   <appender name="STDOUT" class="ch.qos.logback.core.ConsoleAppender">
-    <encoder>
-      <pattern>%d %msg%n</pattern>
+    <encoder class="net.logstash.logback.encoder.LogstashEncoder">
+      <layout class="net.logstash.logback.layout.LogstashLayout">
+        <fieldNames>
+          <timestamp>timestamp</timestamp>
+          <message>msg</message>
+          <logger>lname</logger>
+          <thread>tname</thread>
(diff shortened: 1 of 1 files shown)


ansible/roles/stack-sunbird/templates/learner-service_logback.xml
```diff
@@ -6,8 +6,25 @@
...
+          <timestamp>timestamp</timestamp>
+          <message>msg</message>
+          <logger>lname</logger>
+          <thread>tname</thread>
+          <levelValue>[ignore]</levelValue>
+          <version>[ignore]</version>
+          <stack_trace>exception</stack_trace>
+        </fieldNames>
+        <throwableConverter class="net.logstash.logback.stacktrace.ShortenedThrowableConverter">
+          <maxDepthPerThrowable>30</maxDepthPerThrowable>
```
SMYALTAMASH, line 21: can we variablize these values @amit-tarento ?

ansible/roles/stack-sunbird/templates/learner-service_logback.xml
```diff
@@ -6,8 +6,25 @@
...
+          <maxLength>2048</maxLength>
```
SMYALTAMASH, line 22: can we variablize these values @amit-tarento ?

ansible/roles/stack-sunbird/templates/learner-service_logback.xml
amit-tarento: We can , But not required as its just to print exception stacktrace , with limited verbose.

ansible/roles/stack-sunbird/templates/learner-service_logback.xml
amit-tarento: We can , But not required as its just to print exception stacktrace , with limited verbose.

SMYALTAMASH (approved):

PR #1904, opened 2020-09-17
SC-2019 Refactoring user_feed table with jenkin configuration job

Hari-stackroute:


ansible/roles/cassandra-db-update/tasks/main.yml
```diff
@@ -26,6 +26,6 @@
 
 - name: rename backup file
   become: yes
-  command: mv /home/deployer/usr_external_identity.csv /home/deployer/usr_external_identity_backup.csv
+  command: mv /home/deployer/user_feed.csv /home/deployer/user_feed_backup.csv
```
SMYALTAMASH, line 29: @Hari-stackroute there are 2 things here,
1> use ansible module instead of command for mv
2> use ansible variable for /home/deployer instead of hardcoded value as adopters can use different user instead of deployer if required.

SMYALTAMASH: Make sure you add a timestamp instead of having the overhead of deleting the files all the time @Hari-stackroute

Hari-stackroute: @SMYALTAMASH : appended timestamp to the csv. please check.

SMYALTAMASH: Awesome @Hari-stackroute

SMYALTAMASH (approved):

PR #1913, opened 2020-09-21
Summary aggregate - Variable update

manjudr:


ansible/roles/stack-sunbird/templates/sunbird_lms-service.env
```diff
@@ -139,7 +139,9 @@ user_enrolments_response_cache_ttl={{ user_enrolments_response_cache_ttl | defau
 
 ##collection-summary-agg-api
 
-druid_proxy_api_host={{groups['internal-proxy'][0]}}
+druid_proxy_api_host={{groups['raw-broker'][0]}}
```
keshavprasadms: This will fail for adopter if they dont have druid. How do we handle @manjudr ?

keshavprasadms (approved):

PR #2003, opened 2020-10-23
Issue #SB-21135 feat: Updated LMS configuration

PradyumnaNagendra:


ansible/roles/stack-sunbird/templates/sunbird_lms-service.env
```diff
@@ -19,8 +19,8 @@ sunbird_pg_user={{sunbird_pg_user}}
 sunbird_pg_password={{sunbird_pg_password}}
 sunbird_installation={{sunbird_installation}}
 sunbird_analytics_api_base_url={{sunbird_analytics_api_base_url}}
-sunbird_search_service_api_base_url={{sunbird_search_service_api_base_url}}
-ekstep_api_base_url={{sunbird_content_service_api_base_url}}
+sunbird_search_service_api_base_url=http://search-service:9000
+ekstep_api_base_url=http://content-service:9000
```
rjshrjndrn, line 23: make it as configurable. as, if in case we want to override it should be possible. `{{ lms_ekstep_api_base_url | default('http://content-service:9000') }}`

rjshrjndrn (changes requested):

ansible/roles/stack-sunbird/templates/sunbird_lms-service.env
PradyumnaNagendra: Done

rjshrjndrn (approved):

PR #2029, opened 2020-11-06
Issue #TG642 feat:fix logback file

amit-tarento:


ansible/roles/stack-sunbird/templates/learner-service_logback.xml
```diff
@@ -111,10 +111,6 @@
     <appender-ref ref="defaultLoggerAppender" />
   </logger>
 
-  <logger name="org.sunbird" level="INFO">
```
rjshrjndrn, line 114: This file is templated. So I think it'll be add level as a variable. like
`<logger name="org.sunbird" level="{{learner_log_level | default('ERROR')}}">`
So that by default logging level will be error and if we want to override, it'll be possible

rjshrjndrn (changes requested):

PR #2052, opened 2020-11-23
Nginx logging improvements

rjshrjndrn:
Enabled interface for adding random log entries via ansible variables.

Allow logging to capture all entries from nginx which is not properly parsed.

removing istio scrape configs

changing nginx metrics from rate to irate, so that we will get instant values


kubernetes/helm_charts/core/nginx-public-ingress/values.j2
```diff
@@ -932,7 +932,7 @@ nginxconfig: |
                         ' $request_time $upstream_response_time $pipe'
                         ' "$http_referer" "$http_user_agent" "$sb_request_id"'
                         ' "$http_x_device_id" "$http_x_channel_id" "$http_x_app_id"'
-                        ' "$http_x_app_ver" "$http_x_session_id"';
+                        ' "$http_x_app_ver" "$http_x_session_id" {{nginx_additional_log_fields}}';
```
keshavprasadms, line 935: If I dont use this variable, will your below ruby expression work? You may need a dummy placeholder like space or make the parser below to check for field?

keshavprasadms (commented):
Will the parsing work

keshavprasadms (approved):

PR #2092, opened 2020-12-16
Issue sh-21326 fix: discussion forum envs

VenkatKadiveti:


ansible/roles/stack-sunbird/templates/sunbird_discussions-mw-service.env
```diff
@@ -0,0 +1,2 @@
+discussion_forum_token= {{ discussion_forum_token }}
+discussions_middleware= {{ discussions_middleware }}
```
rjshrjndrn, line 2: why do you need this variable ? if you're looking for the same service, you can use localhost.

rjshrjndrn (changes requested):

rjshrjndrn (approved):

PR #2125, opened 2021-01-07
Adding default vaule for nodebb url

rjshrjndrn:


kubernetes/helm_charts/core/nodebb/values.j2
```diff
@@ -17,11 +17,11 @@ resources:
...
 network:
-  port: 4567 
+  port: 4567
 strategy:
   type: RollingUpdate
   maxsurge: {{ nodebb_maxsurge|default('25%') }}
   maxunavailable: {{ nodebb_maxunavailable|default('25%') }}
 
-nodebb_url: {{ sunbird_nodebb_url }}
+nodebb_url: {{ sunbird_nodebb_url | default(proto+'://'+domain_name+'/discussions') }}
```
keshavprasadms, line 26: We dont need a separate variable. Just use `nodebb_url: "{{proto}}://{{domain_name}}/discussions"`

rjshrjndrn: what If I want to override this ? like I'm using a hosted solution.. Rather than hardcoding, it's good to give an interface to override. 
If in case of any wierd reason, If we've to override then fork is the only option.

keshavprasadms: If you are overriding, then don't you think you need something similar in nginx config also for /disuccsions?

rjshrjndrn: Agreed. but that's a hidden url. and the app devs won't have to call that, right ?

keshavprasadms (approved):

PR #2153, opened 2021-01-12
Updating internal service url path

rjshrjndrn:
search
learner
content

### DIFF
--- a/ansible/inventory/env/group_vars/all.yml
+++ b/ansible/inventory/env/group_vars/all.yml
@@ -111,7 +111,7 @@ sunbird_es_host: "{{groups['es']| join(',')}}"
 
 ## Application server configurations
 sunbird_analytics_api_base_url: "http://{{analyticsapi_ip}}:9000"
-sunbird_search_service_api_base_url: "http://{{sunbird_swarm_manager_lb_ip}}:9004"
+sunbird_search_service_api_base_url: "http://search-service.{{namespace}}.svc.cluster.local:9000"
 sunbird_mail_server_host: "{{ mail_server_host }}"  #SMTP server IP. Could be ignored if not mails are to be sent.
 sunbird_mail_server_port: "{{ mail_server_port }}" #SMTP port. Could be ignored if not mails are to be sent.
 sunbird_mail_server_username: "{{ mail_server_username }}"  #SMTP username. Could be ignored if not mails are to be sent.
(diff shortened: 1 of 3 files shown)


ansible/inventory/env/group_vars/all.yml
```diff
@@ -342,7 +342,7 @@ sunbird_device_register_api: "{{proto}}://{{api_proxy_name}}/v3/device/register/
 sunbird_course_batch_notification_signature: sunbird
 
 sunbird_authorization: "{{core_vault_sunbird_api_auth_token}}"
-sunbird_lms_base_url: "http://{{sunbird_swarm_manager_lb_ip}}:8000"
+sunbird_lms_base_url: "http://lms-service.{{namespace}}.svc.cluster.local:9000"
```
keshavprasadms, line 345: Port 8000 is kong. Please confirm once in some env if this is kong or lms.

ansible/roles/keycloak-deploy/tasks/deploy.yml
```diff
@@ -28,7 +28,7 @@
     regexp: "{{ item.var }}.*"
     state: absent
   with_items:
-    - {var: 'sunbird_lms_base_url'}
+    - {var: '__lms_host__'}
```
keshavprasadms, line 31: Same comment comments as above. This looks to be a kong endpoint

private_repo/ansible/inventory/dev/Core/common.yml
```diff
@@ -51,7 +51,7 @@ certs_badge_image: ""                # Image of the certs badge e.g value is "su
 certs_badge_root_org_id: ""          # Root org ID for which the certs will be created e.g value is "0124229474659643111"
 certs_badge_issuer_name: ""          # The issuer of the certs name e.g value is "Government of karnataka"
 certs_badge_issuer_url:  ""          # Issuer organisation official URL e.g value is "https://testdomain.in"
-cert_service_cloud_storage_type: azure  
+cert_service_cloud_storage_type: azure
```
keshavprasadms, line 54: If we know other clouds are not supported, can we move this to all.yml?

private_repo/ansible/inventory/dev/Core/common.yml
```diff
@@ -136,7 +132,7 @@ sunbird_android_app_url: ""             # Playstore url of android app; default
 # NOT MANDATORY VARS
 sunbird_public_storage_account_name: "{{sunbird_azure_public_storage_account_name}}"
 keycloak_api_management_user_email: "admin@sunbird.org"
-jenkins_url:                            # Jenkins urls to be accessed from slaves; eg "http://10.20.0.12:8080/jenkins" 
+jenkins_url:                            # Jenkins urls to be accessed from slaves; eg "http://10.20.0.12:8080/jenkins"
```
keshavprasadms, line 135: We are not using this variable. Can we remove it from here.

ansible/inventory/env/group_vars/all.yml
rjshrjndrn: this was the lb config for swarm. 8000 -> 9000, because we had one internal lb for all routing

private_repo/ansible/inventory/dev/Core/common.yml
rjshrjndrn: don't know about that. Have to check all the options. I think this worth another PR ?

ansible/inventory/env/group_vars/all.yml
keshavprasadms: I don’t think so. We have /api at the end of the variable in some places.

private_repo/ansible/inventory/dev/Core/common.yml
keshavprasadms: Not sure.. maybe we should check dev - Satish Nirmalkar

PR #2159, opened 2021-01-13
DP-1528: updated composite search config

amitpriyadarshi:


SMYALTAMASH:
Invalid PR, 2159 is raised with correct Details.

PR #2174, opened 2021-01-21
Content-service: Updated consistency level to QUORUM

PradyumnaNagendra:


maheshkumargangula (approved):

ansible/roles/stack-sunbird/templates/content-service_application.conf
```diff
@@ -347,9 +347,11 @@ schema.base_path="{{kp_schema_base_path | default('/home/sunbird/content-service
 cassandra {
   lp {
     connection: "{{ lp_cassandra_connection }}"
+    consistency.level: "QUORUM"
```
SMYALTAMASH, line 350: what if we have a single server for adopters?

ansible/roles/stack-sunbird/templates/content-service_application.conf
```diff
@@ -347,9 +347,11 @@ schema.base_path="{{kp_schema_base_path | default('/home/sunbird/content-service
 cassandra {
   lp {
     connection: "{{ lp_cassandra_connection }}"
+    consistency.level: "{{kp_cassandra_consistency_level | defaults('QUORUM') }}"
```
SMYALTAMASH, line 350: can we have an if else check based on the count of the servers?

SMYALTAMASH: {% if groups['cassandra-2'] is defined %}
consistency.level=quorum
{% else %}
consistency.level=
{% endif %}

or on length of the cassandra servers in the group if it is more than 1?

rjshrjndrn: I think rather than a group check, we can do the count if the group.
{{groups[cassandra ] | length}} > 1

SMYALTAMASH: Agree with this point @rjshrjndrn 👍

PR #2204, opened 2021-02-03
Issue #TG-831 feat:configuring cache ttl for learner

amit-tarento:


ansible/roles/stack-sunbird/templates/sunbird_learner-service.env
```diff
@@ -125,3 +125,6 @@ PORTAL_SERVICE_PORT=http://player.{{namespace}}.svc.cluster.local:3000
 form_api_endpoint={{form_api_endpoint | default('/plugin/v1/form/read')}}
 
 
+# Release-3.7.0
+learner_in_memory_cache_ttl={{learner_in_memory_cache_ttl | default(14400)}}
```
keshavprasadms, line 129: @amit-tarento Please change this to 10m as default. For production we will override with 4h

keshavprasadms (approved):

PR #2239, opened 2021-02-15
fixed the keycloak deployment issue

badaltechiepi:
Disable the monit role in keycloak.yaml file.


keshavprasadms:
@badaltechiepi Please can you state why we want to disable monit? If we have monit installed, it will ensure keycloak is started when crashed right ?

badaltechiepi:
@keshavprasadms  It is looking for swarm-agent-for-alertmanager, swarm-agent-for-prometheus and we are not using the swarm anymore. I have also checked other releases like release-3.4.0 and there also it is desabled.

PR #2254, opened 2021-02-17
Adding mail username as apikey

rjshrjndrn:
We're using azure sendgrid apis and the usename is same
"apikey" for ever. So moving it to all.yaml. Anybody can override this with own variable.

### DIFF
--- a/ansible/inventory/env/group_vars/all.yml
+++ b/ansible/inventory/env/group_vars/all.yml
@@ -660,3 +660,8 @@ dp_postgres_username: analytics
 
 # Will enable cassandra cluster if number of cassandra nodes > 1
 cassandra_cluster_size: "{{ groups['cassandra'] | length }}"
+
+# Azure sendgrid mail server apitoken username
+# This value is constant for sendgrid api authentication.
+# If you're using any other mail server provider, override this value in common.yaml.
+mail_server_username: "apikey"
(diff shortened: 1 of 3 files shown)


ansible/inventory/env/group_vars/all.yml
```diff
@@ -660,3 +660,6 @@ dp_postgres_username: analytics
 
 # Will enable cassandra cluster if number of cassandra nodes > 1
 cassandra_cluster_size: "{{ groups['cassandra'] | length }}"
+
+# Azure sendgrid mail server apitoken username
+mail_server_username: "apikey"
```
keshavprasadms, line 665: Why are we defining a default here?

keshavprasadms: If its only the username to avoid template errors, we should have password and host fields also.

rjshrjndrn: this is a constant for sendgrid. That'll be for the default installations. If you want to override you can do it in common.yaml

keshavprasadms: Okay, please update the comment for this as default value used in case of sendgrid

rjshrjndrn: sure

keshavprasadms (approved):

PR #2267, opened 2021-02-19
fix: adding optional helm timeout variable for all plays

keshavprasadms:


kubernetes/ansible/roles/deploy-badger/tasks/main.yml
```diff
@@ -10,7 +10,7 @@
     dest: "{{ chart_path }}/templates/configmap.yaml"
 
 - name: helm install
-  shell: helm upgrade --install --atomic --timeout 10m {{ release_name }} {{ chart_path }} -n {{ namespace }}
+  shell: helm upgrade --install --atomic --timeout {{helm_install_timeout | d(10m)}} {{ release_name }} {{ chart_path }} -n {{ namespace }}
```
rjshrjndrn, line 13: we've to add ''. 10m is a string right ?

rjshrjndrn (changes requested):

rjshrjndrn (approved):

PR #2268, opened 2021-02-19
Mongodb

sivakarthikeyanoptit:
Ansible script for mongodb setup


ansible/roles/mongodb-cluster/templates/mongodb.yaml
```diff
@@ -0,0 +1,170 @@
+---
```
rjshrjndrn, line 1: why are we having a tasks file in templates ?
and what's this supposed to do ?

rjshrjndrn (changes requested):

rjshrjndrn:
Please add a detailed description of what this change supposed to do.

PR #2271, opened 2021-02-19
SC-2213 : Updated the mapping and indices for user to reduce index size

amit-tarento:


rjshrjndrn:
@amit-tarento Can you please put some description regarding what's this PR is for ?

rjshrjndrn (approved):

PR #2279, opened 2021-02-24
Mongodb ansible role

sivakarthikeyanoptit:
for diksha environment


rjshrjndrn:
Sunbird already has a mongo role. We can reuse that.

sivakarthikeyanoptit:
ok rajesh

PR #2280, opened 2021-02-24
sl-assessment

sivakarthikeyanoptit:
Release 3.7.0


ansible/roles/stack-sunbird/templates/sl-assessments.env
```diff
@@ -0,0 +1,151 @@
+
+# Service Config
+HOST={{sl_assessments_host|default("")}}
```
rjshrjndrn, line 3: will the app work, if the variable is empty ? I don't think so

rjshrjndrn (changes requested):

PR #2295, opened 2021-03-03
feat: TG-797, TG-863 Keycloak 7 configuration changes

keshavprasadms:
- Removed unnessary vars from defaults
- Updated `standalone-ha.xml` for Keycloak 7 and other changes
- Proxy forwarding when running behing a proxy on http port
- JDBC ping for clustering, updated node identifiers for proper clustering, updated keycloak init script to run on IP address for both public and private binds instead of running on `0.0.0.0`
- Updated owners as 2 for all caches
- These configurations work on single node / multi-node
- Wait for keycloak to be full functional before exiting jenkins job, done use pkill command
- Updated jboss wait timeout to 900s (15m)
- Purge DB entries of jgroupsping table in case of a Keycloak crash
- Added keycloak 7 migration script


ansible/roles/keycloak-provision/templates/keycloak-service.j2
```diff
@@ -83,7 +83,7 @@ function start_keycloak() {
         # Make sure log files exist and are writable by $PDIUSER first
         touch $LOGOUT $LOGERR
         chown $keycloakuser:$keycloakuser $LOGOUT $LOGERR
-        su - $keycloakuser -c "cd $keycloakpath && (nohup sh ./standalone.sh -b 0.0.0.0 -bprivate=$(hostname -i) --server-config standalone-ha.xml 0<&- 1>>$LOGOUT 2>>$LOGERR &)"
+        su - $keycloakuser -c "cd $keycloakpath && (nohup sh ./standalone.sh -b={{ansible_default_ipv4.address}} -bprivate={{ansible_default_ipv4.address}} --server-config standalone-ha.xml 0<&- 1>>$LOGOUT 2>>$LOGERR &)"
```
rjshrjndrn, line 86: if keycloak behind an LB, it'll fail. Because keycloak want requests for keycloak IP, but the requests are coming to lb IP. Did you check the scenario? I think it's safe to use 0.0.0.0 as this is a private server

deploy/migrate-to-keycloak7.sh
```diff
@@ -0,0 +1,53 @@
+#!/bin/bash
+
+set -eu -o pipefail
+
```
rjshrjndrn, line 4: we've to have a check for mandatory env vars. Else the script will fail midst some operation.

rjshrjndrn (changes requested):

ansible/roles/keycloak-provision/templates/keycloak-service.j2
keshavprasadms: Yes it works behind an LB also. We have tested this in loadtest env. Moreover, in clustering mode, especially with JDBC Ping, we should not run on 0.0.0.0. It will cause occasional issues in cluster communications.

deploy/migrate-to-keycloak7.sh
keshavprasadms: Thats what set -u does right?

ansible/roles/keycloak-provision/templates/keycloak-service.j2
rjshrjndrn: cool.

deploy/migrate-to-keycloak7.sh
rjshrjndrn: :+1:

PR #2303, opened 2021-03-04
Adding the installation steps for pip

badaltechiepi:
1. New steps for pip installation 
2. Change the mvn.3.6. to mvn3.6


deploy/jenkins/jenkins-server-setup.sh
```diff
@@ -18,7 +18,10 @@ apt-get update
 apt-get install -y jenkins=2.263.4
 
 echo -e "\n\e[0;32m${bold}Installating PIP${normal}"
-apt-get install -y python-pip
+udo apt update
```
rjshrjndrn, line 21: s missing from sudo

deploy/jenkins/jenkins-server-setup.sh
```diff
@@ -18,7 +18,10 @@ apt-get update
 apt-get install -y jenkins=2.263.4
 
 echo -e "\n\e[0;32m${bold}Installating PIP${normal}"
-apt-get install -y python-pip
+udo apt update 
+sudo apt install python2
```
rjshrjndrn, line 22: why are we installing  python2 ? this is not required

deploy/jenkins/jenkins-server-setup.sh
```diff
@@ -18,7 +18,10 @@ apt-get update
...
+curl https://bootstrap.pypa.io/2.7/get-pip.py --output get-pip.py
+sudo python2 get-pip.py
```
rjshrjndrn, line 24: why can't we just use `sudo apt install python3-pip` ?

rjshrjndrn (changes requested):

deploy/jenkins/jenkins-server-setup.sh
badaltechiepi: @rjshrjndrn We need PIP latest version. If you do sudo apt install python3-pip then it will install pip version 8X, for which the ansible is not getting installed when we run the jenkins-server-installation script. So we are adding these steps to install the pip latest version 20X.

deploy/jenkins/jenkins-server-setup.sh
badaltechiepi: @rjshrjndrn Ok. I am correcting this now.

PR #2418, opened 2021-04-14
Enabling proxyvalidator to validate APIs based on API spec defined in docs.sunbird.org

SMYALTAMASH:
Please include a summary of the change and which issue is fixed. Please also include relevant motivation and context. List any dependencies that are required for this change.

### Type of change

Please choose appropriate options.

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

### How Has This Been Tested?

Please describe the tests that you ran to verify your changes in the below checkboxes. Provide instructions so we can reproduce. Please also list any relevant details for your test configuration

- [ ] Ran Test A
- [ ] Ran Test B

**Test Configuration**:
* Software versions:
* Hardware versions:

### Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes
- [x] Any dependent changes have been merged and published in downstream modules


kubernetes/helm_charts/core/apimanager/values.j2
```diff
@@ -25,10 +27,22 @@ strategy:
...
+    memory: {{proxyvalidator_mem_req|default('100Mi')}}
+  limits:
+    cpu: {{proxyvalidator_cpu_limit|default('1000m')}}
+    memory: {{proxyvalidator_memory_limit|default('1024Mi')}}
+
 {{ apimanager_liveness_readiness | to_nice_yaml }}
 
 autoscaling:
-  enabled: {{ apimanager_autoscaling_enabled | default('false') }}
+  enabled: {{ apimanager_autoscaling_enabled | default('true') }}
```
gandham-santhosh, line 45: @SMYALTAMASH we are disabling the autoscaling for all other services by default. Any reason why you want to enable it for api-manger?

SMYALTAMASH: disabled it @gandham-santhosh

deploy/jenkins/jobs/Build/jobs/Core/jobs/Proxyvalidator/config.xml
```diff
@@ -0,0 +1,100 @@
...
+&lt;/b&gt;&lt;/font&gt;</description>
+          <defaultValue>https://github.com/project-sunbird/sunbird.org-docs.git</defaultValue>
+          <trim>true</trim>
+        </hudson.model.StringParameterDefinition>
+        <hudson.model.StringParameterDefinition>
+          <name>docs_repo_branch</name>
+          <description>&lt;font style="color:dimgray;font-size:14px;"&gt;&lt;b&gt;
+  &lt;li&gt;Specify the docs repo branch to pull the spec files from!&lt;/li&gt;
+&lt;/b&gt;&lt;/font&gt;</description>
+          <defaultValue>3.8.0</defaultValue>
```
keshavprasadms, line 52: @SMYALTAMASH Can you change this block to how we do for private_branch in deploy jobs so we can pick release version based on `${public_repo_branch}`. Its a Active choice reactive parameter. Or just change the default value to `${public_repo_branch}` to keep it simpler.

keshavprasadms:
@SMYALTAMASH Good effort in writing this code. But I have a different perspective. Why should we modify the API manager charts and jenkins jobs for this? All adopters would also get these changes which are completely irrelavant and unnecessary right. Instead we should be doing something different to achive this. Here is my one solution, please explore other options too if you would like to.
Recommendations:
- We should have a separate jenkins job and chart for proxy validator. Infact we dont even need a chart for my below solution.
- The proxy job can be a downstream job post kong deployment in whichever env we require it to run.
- The proxy job can have option to enable / disable the validator.

Here is an example you can try which the proxy job code can do - Try it out in your local

```
kubectl create deployment --image=nginx nginx -n default
```
Assume the above deployment is out kong deployment

Lets have a file named `patch.yaml` with below contents. Assume this is the proxy validator image.
```
spec:
  template:
    spec:
      containers:
      - name: redis
        image: redis
```

Our proxy validator jenkins job will just do this when triggered
```
kubectl patch deployments.apps -n default nginx --patch= "$(cat patch.yaml)"
```
Voila! Now we have the side car running.

In case we want to disbale the validator via jenkins flag, we can just run the below
```
kubectl patch deployment -n default nginx  --type json -p='[{"op": "remove", "path": "/spec/template/spec/containers/1"}]'
```
Side car removed.

Same way you can patch the service file also.  Think over this approach and let me know!

SMYALTAMASH:
This approach looks different, will try it out.

deploy/jenkins/jobs/Build/jobs/Core/jobs/Proxyvalidator/config.xml
SMYALTAMASH: nope, this doesn't work as docs repo follows a different release convention

SMYALTAMASH:
Have updated the proxyvalidation deployment to use patch instead of injecting in apimanager.

## Features introduced in this PR:
- Enable proxyvalidator to validate the API requests coming to kong and to adhere to the documentation.
- Have a flag to disable the proxyvalidator in case of emergency
- Made kong and proxyvalidator independent of each other using the approach suggested above in the comments.

kubernetes/ansible/roles/proxyvalidator/tasks/main.yml
```diff
@@ -0,0 +1,35 @@
...
+  template:
+    src: "{{item}}.j2"
+    dest: "/tmp/{{item}}"
+    mode: 0755
+  with_items:
+    - proxyvalidatorServiceRevert.yml
+  when: enable_proxyvalidator == "no"
+
+- name: delete the proxyvalidator container
+  shell: "kubectl patch deployment -n \"{{namespace}}\" apimanager --type json -p='[{\"op\": \"remove\", \"path\": \"/spec/template/spec/containers/0\"}]'"
```
keshavprasadms, line 30: Can yo confiirm the `path`? `0` will be kong pod and `1` will be proxy pod. Please check it out

SMYALTAMASH: yes path will be 0 for patched container. it is correct.

kubernetes/helm_charts/core/apimanager/templates/deployment.yaml
```diff
@@ -52,6 +52,17 @@ spec:
               command:
               - kong
               - quit
+{{- if .Values.enable_apiproxyvalidator }}
```
keshavprasadms, line 55: Do we still need these in this chart? Please can you revert them to original state?

SMYALTAMASH: just for additional check, do i remove them?

SMYALTAMASH: we don't need them, will remove it right away.

keshavprasadms (changes requested):
Looks much better. Please take a look at the comments. Also do try out running multiple times with same option like `yes` or `no` to ensure we are not creating incorrect configs.

kubernetes/helm_charts/core/apimanager/templates/deployment.yaml
SMYALTAMASH: removed it.

SMYALTAMASH:
Have tried with multiple iterations, if we run with no multiple times the job fails as it can't remove the only container which is apimanager in our case, it is working as expected.

keshavprasadms (approved):

keshavprasadms:
Nice work!

PR #2497, opened 2021-05-04
feat: TG-957 adding mongo import export

keshavprasadms:
- Adding import via json option for graylog
- Using this approach we can inject ansible variables specific to env
- Export the data using `mongoexport --db=graylog --collection=collection_name --out=file_name.json`
- Import the data using `mongoimport --drop --db=graylog --collection=collection_name --file=file_name.json`
- Any custom / user made changes on the UI will be removed when this job is run
- To make the change permanent, a PR with the configuration has to be sent to this role
- If a user wishes not to lose their manual changes, then the will need to define this variable in private repo `graylog_mongo_collections_drop: false`, but this may cause issues if new data from repo and manual changes have the same primary key in mongo db (rare case)


rjshrjndrn (changes requested):
can't we keep all these jsons in files/ directory, as there are no templating?

keshavprasadms:
> can't we keep all these jsons in files/ directory, as there are no templating?

@rjshrjndrn There is templating. As of now its only in one file, in future we will have more templating coming - https://github.com/project-sunbird/sunbird-devops/pull/2497/files#diff-b881b0e4f2e16a40fbe3958b4c3a6dc21098510ff9048c940b7bc5ecc683aa67R2

rjshrjndrn (approved):

PR #2500, opened 2021-05-05
feat: TG-957 exposing graylog via nginx location

keshavprasadms:
- Enabling graylog location via nginx in the route `{{ domain }}/graylog/` using an optional variable
- When the variable `graylog_open_to_public` is defined and set to `true` in private, graylog will be open to public
- User will be auto logged in as `viewer` as we are passing the username as a header
- To login as `admin`, user should tunnel to the graylog VM or use VPN
- Adding Graylog service and external endpoints for filebeat helm chart
- `graylog_trusted_proxies` must be set in private repo in order to make graylog work with reverse proxy
- Few variables renamed, indentations for ansible variables for better readibility
- Cleaned up unnecessary files


kubernetes/helm_charts/logging/filebeat/templates/graylog-service.yaml
```diff
@@ -0,0 +1,32 @@
+---
+apiVersion: v1
```
rjshrjndrn, line 2: Nice One !!

rjshrjndrn (approved):

PR #2504, opened 2021-05-05
feat: TG-957 option to use graylog behind a private LB with Oauth

keshavprasadms:
`send_logs_to_graylog` defines if the logs need to be sent to graylog or elasticsearch (EFK stack)
`graylog_open_to_public` defines whether graylog is accessible via domain
`graylog_open_to_private` defines if graylog needs to accessed only via internal network via a private LB
Using the above 3 variables, you can get exactly what you want

Below are some sensible scenarios -
1. You want the EFK stack. Don't define any of the above 3 variables
2. You just want graylog, you dont care how you will check the logs (tunneling maybe). In this case just set  `send_logs_to_graylog: true`
3. You want graylog to be available via domain, then set `graylog_open_to_public: true` along with `send_logs_to_graylog: true`
4. You want graylog but you want to access it only via private LB, then set `graylog_open_to_private: true` and `send_logs_to_graylog: true`
5. You made a blunder and set all of them as `graylog_open_to_private: true`, `graylog_open_to_public: true`, `send_logs_to_graylog: true`. In this case you will end up with `Scenario 3`
6. For private LB, you need to pre-define a private ip in your private repo using the variable `oauth2_proxy_lb_ip`. This is a pre-requisite so we can template the `redirect_url` with `oauth2_proxy_lb_ip.nip.io`
7. Ensure you add the `http://oauth2_proxy_lb_ip.nip.io` in google oauth admin console as a whitelisted URI and a valid redirect as `http://oauth2_proxy_lb_ip.nip.io/oauth2/callback`

### DIFF
--- a/kubernetes/ansible/roles/logging/templates/oauth2-proxy.yaml
+++ b/kubernetes/ansible/roles/logging/templates/oauth2-proxy.yaml
@@ -8,13 +8,7 @@ config:
 {% else %}
     redirect_url = "{{ oauth2_proxy_redirect_url }}"
 {% endif %}
-{% if send_logs_to_graylog is defined and send_logs_to_graylog is true %}
-    upstreams = "http://graylog.logging.svc.cluster.local"
-{% else %}
-    upstreams = "http://{{ kibana_service }}"
-{% endif %}
     email_domains = {{ kibana_oauth_authenticated_email_domains_kube }}
-    authenticated_emails_file = "/etc/oauth2/auth-emails/authenticated_email.txt"
     cookie_expire = "24h"
     cookie_refresh = "4h"
 {% if (graylog_open_to_public is defined and graylog_open_to_public is true) or (send_logs_to_graylog is not defined or send_logs_to_graylog is false) %}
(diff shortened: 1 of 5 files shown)


kubernetes/ansible/roles/logging/templates/oauth2-proxy.yaml
```diff
@@ -41,3 +50,20 @@ authenticatedEmailsFile:
      {% endfor %}
 
 fullnameOverride: "oauth2-proxy"
+
+{% if (graylog_open_to_public is not defined or graylog_open_to_public is false) and (send_logs_to_graylog is defined and send_logs_to_graylog is true) and (graylog_open_to_private is defined and graylog_open_to_private is true) %}
+service:
+  type: LoadBalancer
+  loadBalancerIP: {{ oauth2_proxy_lb_ip }}
+  port: 80
+{% if nginx_private_ingress_annotations is defined and nginx_private_ingress_annotations %}
```
rjshrjndrn, line 59: we don't need this if condition here, as it's mandatory to give the annotation for private LB.

keshavprasadms: If the user has already defined annotation for their private nginx, we can reuse the same variable here right? Else we wont be able to override in private as azure is hardcoded in the next section.

kubernetes/ansible/roles/logging/templates/oauth2-proxy.yaml
```diff
@@ -41,3 +50,20 @@ authenticatedEmailsFile:
...
+# you can define annotations in ansible for custom behaviour
+# eg:
+# nginx_private_ingress_annotations:
+#   cloud.google.com/load-balancer-type: "Internal"
+  annotations: {{nginx_private_ingress_annotations|to_json}}
+{% else %}
+  annotations:
+    service.beta.kubernetes.io/azure-load-balancer-internal: "true"
```
rjshrjndrn, line 101: Please remove this, as it'll only work in azure. All the other providers will ignore this and will create public LB. It's better to error out than an undesired behaviour.

kubernetes/ansible/roles/logging/templates/oauth2-proxy.yaml
rjshrjndrn: :point_down:

kubernetes/ansible/roles/logging/templates/oauth2-proxy.yaml
keshavprasadms: Done. Reusing nginx variable to not introduce another variable.

rjshrjndrn (approved):

PR #2663, opened 2021-07-01
SB-24483 : Feat: Env variable change

kiranharidas187:
<table>
<td>
  <a href="https://project-sunbird.atlassian.net/browse/SB-24483" title="SB-24483" target="_blank">EpicSB-24483</a>  Observation Consumption stories for Manage Learn
  </td></table>
  <br />

### DIFF
--- a/ansible/roles/stack-sunbird/templates/sunbird_player.env
+++ b/ansible/roles/stack-sunbird/templates/sunbird_player.env
@@ -156,3 +156,4 @@ sunbird_portal_session_secret={{sunbird_portal_session_secret}}
 
 #release-4.0.0
 sunbird_base_proto={{sunbird_base_proto | default(proto)}}
+ML_SERVICE_BASE_URL={{ML_SERVICE_BASE_URL}}
--- a/ansible/roles/stack-sunbird/templates/sunbird_player.env
+++ b/ansible/roles/stack-sunbird/templates/sunbird_player.env
@@ -156,4 +156,4 @@ sunbird_portal_session_secret={{sunbird_portal_session_secret}}
 
 #release-4.0.0
 sunbird_base_proto={{sunbird_base_proto | default(proto)}}
-ML_SERVICE_BASE_URL={{ML_SERVICE_BASE_URL}}
+ML_SERVICE_BASE_URL={{ml_server_url}}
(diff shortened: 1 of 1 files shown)


ansible/roles/stack-sunbird/templates/sunbird_player.env
```diff
@@ -156,3 +156,4 @@ sunbird_portal_session_secret={{sunbird_portal_session_secret}}
 
 #release-4.0.0
 sunbird_base_proto={{sunbird_base_proto | default(proto)}}
+ML_SERVICE_BASE_URL={{ML_SERVICE_BASE_URL}}
```
rjshrjndrn, line 159: Keep the right hand side variable as small letters. And what's the use of the variable? I meant the variable name is not contextful.

rjshrjndrn (changes requested):

ansible/roles/stack-sunbird/templates/sunbird_player.env
kiranharidas187: @rjshrjndrn Right side variable change done. This URL is specific to ML backend service and this variable will be removed by 4.2 release.

rjshrjndrn: Which backend service specifically? 
Means ml has multiple services running, like project, reports, etc.
So if it's pointing to a specific service then it'll be better to have variables like `{{ml_project_service_url | default('project-service:3000'}}`

The default value because, project service is running in Kubernetes, and the private endpoint for that service not gonna change.

ansible/roles/stack-sunbird/templates/sunbird_player.env
```diff
@@ -156,3 +156,4 @@ sunbird_portal_session_secret={{sunbird_portal_session_secret}}
 
 #release-4.0.0
 sunbird_base_proto={{sunbird_base_proto | default(proto)}}
+ML_SERVICE_BASE_URL={{ml_server_url}}
```
rjshrjndrn, line 159:
```
{{ml_survey_url|default(proto+'://' +domain_name) }}
```

rjshrjndrn (changes requested):

rjshrjndrn (approved):

PR #2791, opened 2021-08-10
UCI Nginx Config - Remove rewrite rules

ChakshuGautam:


ChakshuGautam:
@keshavprasadms please review.

kubernetes/helm_charts/core/nginx-private-ingress/templates/configmap.yaml
```diff
@@ -174,17 +174,17 @@ data:
         }
         location /gql-service/ {
           set $target http://gql-service.{{ .Values.namespace }}.svc.cluster.local:8080;
-          rewrite ^/gql-service/(.*) /$1 break;
+          rewrite ^/(.*) /$1 break;
```
rjshrjndrn, line 177: I don't think we need this rewrite block, as we're replacing the same string. like, replace apple with apple. So it'll be good to drop all these rewrites.

rjshrjndrn (changes requested):

kubernetes/helm_charts/core/nginx-private-ingress/templates/configmap.yaml
ChakshuGautam: Great thanks. Will just update this.

ChakshuGautam:
@rjshrjndrn Please check.

keshavprasadms (approved):

ChakshuGautam:
@keshavprasadms @rjshrjndrn Currently the `http://11.2.6.6/gql-service/console/` is getting translated to `/gql-service/console/` but I need it to be `/console/`. What changed do I need to make this routing work?

PR #2957, opened 2021-10-18
UCI Config Changes

ChakshuGautam:
Changes

1. Update Kafka topic name convention from `com.*` to `*`
2. Add `FUSIONAUTH_KEY`and `FUSIONAUTH_URL` to inbound, orchestrator environments.
3. Fix `uci_api` service to `uci`.
4. Change liveness and readiness for transformer since it takes longer to start the service.


ansible/roles/stack-sunbird/defaults/main.yml
```diff
@@ -875,23 +875,26 @@ uci_outbound_liveness_readiness:
...
+  livenessProbe:
+    failureThreshold: 2
     httpGet:
-      path: /service/health
+      path: /health
       port: 9091
-    initialDelaySeconds: 15
-    periodSeconds: 15
+      scheme: HTTP
+    initialDelaySeconds: 90
```
keshavprasadms, line 884: Any reason why the initial delay needs to be so long? Usually we keep it short like 15s including the periodic probing

ChakshuGautam: Yes. The service downloads 200-300 small files from a different server and caches it on its end. That takes around 30-40 seconds. I have kept a buffer since the number of files will keep increasing.

ChakshuGautam: Different service = ODK service.

PR #2973, opened 2021-10-21
SB-26849 : Add updateable env variable enable tenant config

sknirmalkar89:
This will enable group notifications for a  user belonging to different states.
The string values :  

“*” : It will allow  for all tenants

“tn”: it will allow  only for tamilnadu

“OFF“: To disable for all states

To allow multiple channel . please write as comma separated : e.g : tn,cg  value

### DIFF
--- a/ansible/roles/stack-sunbird/templates/sunbird_groups-service.env
+++ b/ansible/roles/stack-sunbird/templates/sunbird_groups-service.env
@@ -35,4 +35,4 @@ groups_redis_ttl={{groups_redis_ttl | default(86400)}}
 user_redis_ttl={{user_redis_ttl | default(3600)}}
 notification_service_base_url=http://notification-service:9000
 notification_service_api_url=/v2/notification/send
-enable_tenant_config=*
+enable_tenant_config={{enable_tenant_config| default('*')}}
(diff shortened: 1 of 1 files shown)


ansible/roles/stack-sunbird/templates/sunbird_groups-service.env
```diff
@@ -35,4 +35,4 @@ groups_redis_ttl={{groups_redis_ttl | default(86400)}}
 user_redis_ttl={{user_redis_ttl | default(3600)}}
 notification_service_base_url=http://notification-service:9000
 notification_service_api_url=/v2/notification/send
-enable_tenant_config=*
+enable_tenant_config={{enable_tenant_config| default('*')}}
```
rjshrjndrn, line 38: what's the meaning of this * ? can you please add a comment above what should be the variable value, and for what functionality we're using this?

sknirmalkar89: Updated please check

rjshrjndrn: where did you udpate? what I meant was https://github.com/project-sunbird/sunbird-devops/blob/7e422a9ab8b6af53c0387822467d80edbaf0e783/ansible/roles/stack-sunbird/templates/sunbird_player.env#L139-L143

PR #2997, opened 2021-11-08
UCI: Add Cache Layer

ChakshuGautam:
API request caching for UCI APIs.

Caveats
1. Does not have HPA. Can only be scaled vertically. Will suffice the current (and future) needs of UCI. Load tested this at  16k requests per second.
2. Will require a higher CPU (1000m) at production.
3. Will require a higher memory (1000MB) at production.


ChakshuGautam:
@keshavprasadms Please review.

keshavprasadms:
@rjshrjndrn Please review this

rjshrjndrn:
Hi @ChakshuGautam why do we need a separate caching mechanism? Can't we just use nginx?

PR #3179, opened 2022-02-08
transformer env file updates

surabhi-mahawar:
New env variable for telemetry event context pdata id

### DIFF
--- a/ansible/roles/stack-sunbird/templates/transformer.env
+++ b/ansible/roles/stack-sunbird/templates/transformer.env
@@ -50,3 +50,6 @@ ASSESSMENT_ONE_LEVEL_UP_CHAR={{uci_one_level_up_char | default("#") }}
 NETCORE_WHATSAPP_AUTH_TOKEN={{uci_netcore_whatsapp_token}}
 NETCORE_WHATSAPP_SOURCE={{uci_netcore_whatsapp_source}}
 NETCORE_WHATSAPP_URI={{uci_netcore_whatsapp_uri | default('https://waapi.pepipost.com/api/v2/')}}
+
+#Telemetry
+TELEMETRY_EVENT_PDATA_ID={{uci_telemetry_event_pdata_id | default('dev.uci.diksha')}}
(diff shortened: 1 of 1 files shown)


ansible/roles/stack-sunbird/templates/transformer.env
```diff
@@ -49,4 +49,7 @@ ASSESSMENT_ONE_LEVEL_UP_CHAR={{uci_one_level_up_char | default("#") }}
 #Netcore
 NETCORE_WHATSAPP_AUTH_TOKEN={{uci_netcore_whatsapp_token}}
 NETCORE_WHATSAPP_SOURCE={{uci_netcore_whatsapp_source}}
-NETCORE_WHATSAPP_URI={{uci_netcore_whatsapp_uri | default('https://waapi.pepipost.com/api/v2/')}}
\ No newline at end of file
+NETCORE_WHATSAPP_URI={{uci_netcore_whatsapp_uri | default('https://waapi.pepipost.com/api/v2/')}}
+
+#Telemetry
+TELEMETRY_EVENT_PDATA_ID={{uci_telemetry_event_pdata_id | default('dev.uci.diksha')}}
```
keshavprasadms, line 55: Please remove references of diksha. This is sunbird repo.

keshavprasadms (changes requested):
Make the changes please

ansible/roles/stack-sunbird/templates/transformer.env
keshavprasadms: Also this will always be `dev.uci` in all env. Refer to other code for such values and it needs to be prefixed with `{{env}}`

surabhi-mahawar:
@keshavprasadms I have removed the default value, as we don't really need it, the values from env will work.

ansible/roles/stack-sunbird/templates/transformer.env
ChakshuGautam: @keshavprasadms we need something like this `{{env}}.uci.{{sunbird/disksha depending on dev/staging/prod/preprod}}`. Can you suggest how we can get `instance` of deployment based on env. thanks.

keshavprasadms: {{env}} will give you data like, dev, staging, prod etc. There is another variable called `sunbird_instance` which can have data like sunbird / diksha / vidyadaan etc.

keshavprasadms (approved):

keshavprasadms (approved):

PR #3186, opened 2022-02-11
Env changes for redis

surabhi-mahawar:
Environment variables for redis server & Service environment

### DIFF
--- a/ansible/roles/stack-sunbird/templates/inbound.env
+++ b/ansible/roles/stack-sunbird/templates/inbound.env
@@ -42,3 +42,11 @@ FUSIONAUTH_URL=http://fusionauth-service.{{namespace}}.svc.cluster.local:9011
 FUSIONAUTH_KEY={{fusionauth_service_admin_key}}
 
 CAMPAIGN_ADMIN_TOKEN={{uci_api_admin_token}}
+
+#Environment
+ENV={{env}}
+
+#Redis
+REDIS_HOST={{redis_host}}
+REDIS_PORT={{redis_port}}
+REDIS_PASS={{redis_pass}}
--- a/ansible/roles/stack-sunbird/templates/inbound.env
+++ b/ansible/roles/stack-sunbird/templates/inbound.env
@@ -47,6 +47,7 @@ CAMPAIGN_ADMIN_TOKEN={{uci_api_admin_token}}
 ENV={{env}}
 
 #Redis
-REDIS_HOST={{redis_host}}
-REDIS_PORT={{redis_port}}
-REDIS_PASS={{redis_pass}}
+REDIS_HOST={{sunbird_redis_host}}
+REDIS_PASS=""
+REDIS_PORT=6379
+REDIS_DB_INDEX={{redis_db_index_uci | default('6')}}
--- a/ansible/roles/stack-sunbird/templates/inbound.env
+++ b/ansible/roles/stack-sunbird/templates/inbound.env
@@ -48,6 +48,6 @@ ENV={{env}}
 
 #Redis
 REDIS_HOST={{sunbird_redis_host}}
-REDIS_PASS=""
+REDIS_PASS={{sunbird_redis_pass | default('')}}
 REDIS_PORT={{sunbird_redis_port | default(6379)}}
-REDIS_DB_INDEX={{redis_db_index_uci | default('6')}}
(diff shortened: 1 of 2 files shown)


ansible/roles/stack-sunbird/templates/inbound.env
```diff
@@ -42,3 +42,11 @@ FUSIONAUTH_URL=http://fusionauth-service.{{namespace}}.svc.cluster.local:9011
 FUSIONAUTH_KEY={{fusionauth_service_admin_key}}
 
 CAMPAIGN_ADMIN_TOKEN={{uci_api_admin_token}}
+
+#Environment
+ENV={{env}}
+
+#Redis
+REDIS_HOST={{redis_host}}
```
keshavprasadms, line 50: What are the values for these? Where are the default values defined? What about the redis index?

surabhi-mahawar: I have updated the default values in the sheet.

https://docs.google.com/spreadsheets/d/1xoOXNtDR7fnW9CU4mj8Q2xZr2mkdqpXdAfCOjD9BfrY/edit#gid=965973714

keshavprasadms: No, these cannot be added in private repo. We already have a ton of redis instance variables pre-defined. Please reuse them. We cannot be introducing new variables for each service.

keshavprasadms: Example - https://github.com/project-sunbird/sunbird-devops/blob/release-4.7.0/ansible/roles/stack-sunbird/templates/sunbird_learner-service.env#L98-L99

keshavprasadms: You also need to decide which redis instance you want to use. We have multiple of them. Example of another one - https://github.com/project-sunbird/sunbird-devops/blob/release-4.7.0/ansible/roles/stack-sunbird/templates/dial-service_application.conf#L171

keshavprasadms: And what redis db index you are planning to write? You cannot be writing in DB index 0 as we already have other applications writing data. So you need to keep yourself isolated unless you are also writing similar data to same index.

ansible/roles/stack-sunbird/templates/inbound.env
```diff
@@ -49,5 +49,5 @@ ENV={{env}}
 #Redis
 REDIS_HOST={{sunbird_redis_host}}
 REDIS_PASS=""
-REDIS_PORT=6379
+REDIS_PORT={{sunbird_redis_port | default(6379)}}
 REDIS_DB_INDEX={{redis_db_index_uci | default('6')}}
```
keshavprasadms, line 53: Please change the default index to 7

ansible/roles/stack-sunbird/templates/inbound.env
```diff
@@ -49,5 +49,5 @@ ENV={{env}}
 #Redis
 REDIS_HOST={{sunbird_redis_host}}
 REDIS_PASS=""
```
keshavprasadms, line 51: Please add a variable as a placeholder like this - https://github.com/project-sunbird/sunbird-devops/blob/master/ansible/roles/stack-sunbird/templates/sunbird_nodebb.env#L11

keshavprasadms (changes requested):
@surabhi-mahawar Please make the required changes

ansible/roles/stack-sunbird/templates/inbound.env
surabhi-mahawar: is there a variable available for redis password like "sunbird_redis_pass", or should I just add a new one here. 
Something like this should work right?
REDIS_PASS={{sunbird_redis_pass | default('')}}

ansible/roles/stack-sunbird/templates/inbound.env
surabhi-mahawar: is there a variable available for redis password like "sunbird_redis_pass", or should I just add a new one here.
Something like this should work right?
REDIS_PASS={{sunbird_redis_pass | default('')}}

surabhi-mahawar:
@keshavprasadms I have made the changes, Please review it.

keshavprasadms (approved):

PR #3219, opened 2022-03-11
Azure Blob Storage Env Variables

surabhi-mahawar:


keshavprasadms (changes requested):
No unnecessary new varibles should be introduced. Please refer to existing env files and reuse them. For example - https://github.com/project-sunbird/sunbird-devops/blob/master/ansible/roles/stack-sunbird/templates/sunbird_player.env#L46-L47
Any URL connections string, please try to derive it from existing varibles if you can.
For example - https://github.com/project-sunbird/sunbird-devops/blob/master/ansible/inventory/env/group_vars/all.yml#L39

ansible/roles/stack-sunbird/templates/inbound.env
```diff
@@ -51,3 +51,8 @@ REDIS_HOST={{sunbird_redis_host}}
 REDIS_PASS={{sunbird_redis_pass | default('')}}
 REDIS_PORT={{sunbird_redis_port | default(6379)}}
 REDIS_DB_INDEX={{redis_db_index_uci | default('7')}}
+
+#Azure Config
+AZURE_BLOB_STORE_CONTAINER={{sunbird_azure_uci_container_name}}
```
keshavprasadms, line 56: Please add a default value like below so that we can avoid adding additonal variables
`{{sunbird_azure_uci_container_name | default('uci-' + env )}}`

keshavprasadms (approved):

PR #3221, opened 2022-03-14
Netcore Env

surabhi-mahawar:
Netcore Environment Variables for Media Capture

### DIFF
--- a/ansible/roles/stack-sunbird/templates/inbound.env
+++ b/ansible/roles/stack-sunbird/templates/inbound.env
@@ -56,3 +56,9 @@ REDIS_DB_INDEX={{redis_db_index_uci | default('7')}}
 AZURE_BLOB_STORE_CONTAINER={{sunbird_azure_uci_container_name | default('uci-' + env )}}
 AZURE_BLOB_STORE_ACCOUNT_KEY={{sunbird_private_storage_account_name}}
 AZURE_BLOB_STORE_ACCOUNT_NAME={{sunbird_private_storage_account_key}}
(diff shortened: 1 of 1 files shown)


ansible/roles/stack-sunbird/templates/inbound.env
```diff
@@ -56,3 +56,9 @@ REDIS_DB_INDEX={{redis_db_index_uci | default('7')}}
 AZURE_BLOB_STORE_CONTAINER={{sunbird_azure_uci_container_name | default('uci-' + env )}}
 AZURE_BLOB_STORE_ACCOUNT_KEY={{sunbird_private_storage_account_name}}
 AZURE_BLOB_STORE_ACCOUNT_NAME={{sunbird_private_storage_account_key}}
+
+#Netcore
+NETCORE_WHATSAPP_AUTH_TOKEN={{uci_netcore_whatsapp_token}}
```
keshavprasadms, line 61: Add a comment on what are these variables for and how to generate values for them etc.

surabhi-mahawar: We already have these variables in outbound environment, we want to add these to inbound as well.
https://github.com/project-sunbird/sunbird-devops/blob/release-4.8.0/ansible/roles/stack-sunbird/templates/outbound.env

keshavprasadms: ok

keshavprasadms (approved):

PR #3510, opened 2022-09-09
feat: SB-30654 grouping azure specific tasks into a single role

keshavprasadms:
<table>
<td>
  <a href="https://project-sunbird.atlassian.net/browse/SB-30654" title="SB-30654" target="_blank">StorySB-30654</a>  CSP - Ansible Azure Task Changes and Variable Generilization
  </td></table>
  <br />
 

- This PR adds a role named `azure-cloud-storage` which includes all the commonly used azure storage related tasks.
- All the other plays and tasks have been updated to invoke the `azure-cloud-storage` role where ever required instead of using the inline `az` command line statements.
- The `azure-cloud-storage` will be invoked when when the value of the variable `cloud_service_provider` equals `azure`.
- The plays / tasks can be extended by adding other CSP related tasks in a similar fashion.
- New / existing variables added / modified based on requirements.
- Installing specific azure cli package version
- Deleted unnecessary roles / plays
- Updated private repo templates
- Updated vars to use generic naming convention to reuse across cloud providers
- Updated `UploadFAQs` and `Upload_RC_Schema` jenkins job `config.xml` to use generic named parameters
- Added new or missing jenkins job `config.xml`


santhosh-tg:
As discussed , it would be good to have a common variable name  instead of cloud specific one in the general roles 
ex: https://github.com/project-sunbird/sunbird-devops/blob/release-5.1.0/ansible/roles/postgresql-backup/defaults/main.yml#L3
we can replace it with **postgresql_backup_storage**

There might be other places also where we can make this change.

G33tha (approved):

PR #3915, opened 2023-10-31
ED-3079: deploy nginx to support cors for s3 compatible object storage

ddevadat:
<table>
<td>
  <a href="https://project-sunbird.atlassian.net/browse/ED-3079" title="ED-3079" target="_blank">BugED-3079</a>  deploy nginx to support cors for s3 compatible object storage
  </td></table>
  <br />
 

OCI s3 compatible url doesnt support CORS.

While using s3 compatible url in the form of https://<namespace>.compat.objectstorage.<region>.oraclecloud.com works in most of the case, it will not work where CORS is involved. So any AJAX call going from player will fail.

 

In order to provide appropriate CORS response, there should be something in place when using S3 Compatible API URL with a Web Application (that is the Sunbird-ED portal, aka player) which required AJAX requests. It can be CDN, LB, RP or combination of them. 

a new proxy config has been added to public ingress chart. This is a conditional block only for oci

once public-nginx-ingress is created,  Use this public ip to add a dns zone record with the name 
files.{{domain_name}}

e.g if your domain_name is ed.sunbird.in, then the record should files.ed.sunbird.in


ddevadat:
@santhosh-tg updated the PR.
In this approach instead of creating a new deployment and lb, added the proxy config for cors in the existing nginx-public-ingress.  Please review if this works.

gandham-santhosh (approved):

